PR #2, opened 2019-09-17
Some bug fix in line with compiler complains

roshii:
I am begging with C++ and listening to what the compiler says.


stepansnigirev:
Thanks!

PR #5, opened 2019-09-22
uBitcoin version update

miketlk:
This update of uBitcoin.lib is required to instruct mbed fetching newer version of external library. Otherwise project is not built due to missing HDPrivateKey::fingerprint() overridden method and other incompatibilities.


stepansnigirev:
Thanks! Missed it somehow...

PR #28, opened 2019-12-23
implement factory settings

gorazdko:
## Abstract

Device is reset to factory settings after 10 failed pin attempts; https://github.com/cryptoadvance/specter-diy/pull/26#issuecomment-568223353

### DIFF
--- a/src/pin.py
+++ b/src/pin.py
@@ -4,9 +4,14 @@ from bitcoin import bip39
 from platform import storage_root
 from rng import get_random_bytes
 from ubinascii import hexlify, unhexlify
+from micropython import const
+from gui.popups import alert
 import ujson as json
+import os
+import sys
 
 secret_fname = "%s/%s" % (storage_root, "secret.json")
+login_fname = "%s/%s" % (storage_root, "login.json")
 
 class Key:
     key = None
@@ -44,6 +49,9 @@ class Secret:
                      hexlify(Secret.hmac).decode('utf-8')))
 
 class Pin:
+    ATTEMPTS_MAX = const(10)
+    counter = ATTEMPTS_MAX
+
     @staticmethod
     def is_pin_valid():
         assert(Secret.hmac != None)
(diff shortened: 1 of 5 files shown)


src/pin.py
```diff
@@ -4,9 +4,12 @@
 from platform import storage_root
 from rng import get_random_bytes
 from ubinascii import hexlify, unhexlify
+from micropython import const
 import ujson as json
+import os
 
 secret_fname = "%s/%s" % (storage_root, "secret.json")
+login_fname = "%s/%s" % (storage_root, "login.json")
```
gorazdko: i used a new file for the pin counter, as using secret_fname would require editing it, which doesnt seem optimal to me (e.g. the file could get corrupt?). Let me know if you think otherwise and i will corect it.

src/pin.py
```diff
@@ -52,6 +58,35 @@ def is_pin_valid():
...
+        os.remove(login_fname)
+
+def factory_settings():
+    # unlike listdir, ilistdir is supported by unix and stm32 platform
+    files = os.ilistdir(storage_root)
+    for file in files:
+        # regular file
+        if file[1] == 0x8000:
+             f = "%s/%s" % (storage_root, file[0])
+             os.remove(f)
```
gorazdko: should we try catch these? If yes, what do we do in except sentence?

stepansnigirev: I think we should catch it and show an alert that factory reset did not work for some reason.

Also, I am thinking about folders in the `storage_root` - we have wallet files there, so if we call it a factory reset we should delete the content of the `storage_root` recursively.

src/pin.py
stepansnigirev: I think it is a good approach.
Later, when we move the secret key to a fixed location in flash, we should move the PIN counter away from micropython storage as well.

stepansnigirev:
Thank you! Sorry for a delay with the review - Christmas, the New Year, relatives...

One small question to discuss - should we delete wallet files or not when we do a factory reset?

What about the touchscreen calibration file? It's not there at the moment as I still didn't implement it, but would it make sense to keep it or better to remove everything and go to factory reset?

gorazdko:
Wishing you all the best in 2020!

Yes, i will delete it recursively, and add a blacklist array to exempt folders/files. For now we can have an empty array.
I will have to traverse the filesystem though since upython supports deleting individual file only, i think. But I will check tomorrow again.

gorazdko:
force pushed changes

stepansnigirev:
Great! Now we have a fully functional PIN code!

PR #53, opened 2020-03-03
qr scanner: animated qr

### LINKED ISSUE #47: Implement scanning of animated QR code
The problem:
Multisig wallet descriptors and transactions are pretty large. It's hard to scan even 2 of 3 multisig transaction with more than one input.

The idea is to split QR code on software and animate it. We can put a part number in the beginning of the QR code for example. Such that if we see QR code of the form `p2of5 <data>` we know that we scanned 2nd of 5 QR codes.
`QRScanner` class should be able to detect these animated QR codes and continue scanning until we got all the codes. Then it should combine them and send full data.

gorazdko:
## Abstract

Close #47

## Status

* ready for review
* tested on stm32 platform (OK)
* does not work in simulator over telnet (i might open an issue for that)

> QRScanner class should be able to detect these animated QR codes and continue scanning until we got all the codes

Currently, the user must press the button when he is ready to scan the next code. Is it possible to achieve this without a button, automatically?

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -114,7 +114,7 @@ def add_new_wallet():
                           "Scanning.. Click \"Cancel\" to stop.",
                           callback=cancel_scan)
     gui.update(30)
-    qr_scanner.start_scan(parse_new_wallet)
+    qr_scanner.start_scan(parse_new_wallet, qr_animated, add_new_wallet)
 
 @catchit
 def wallets_menu():
@@ -235,13 +235,16 @@ def parse_transaction(b64_tx, success_callback=None, error_callback=None):
         cancel=cb_with_args(error_callback, "user cancel")
     )
 
+def qr_animated(indx, callback):
+    gui.alert("Animated QR: %s/%s" % (indx[0], indx[1]), "Proceed to scanning code No. %s" % str(indx[0]+1), callback)
+
 @catchit
 def scan_transaction():
     screens.show_progress("Scan transaction to sign",
                           "Scanning.. Click \"Cancel\" to stop.",
                           callback=cancel_scan)
     gui.update(30)
-    qr_scanner.start_scan(parse_transaction)
+    qr_scanner.start_scan(parse_transaction, qr_animated, scan_transaction)
 
 @catchit
 def verify_address(s):
@@ -607,9 +610,13 @@ def host_callback(data):
     # - signmessage <message>
     # - showdescraddr <descriptor>
 
+def qr_scanner_error(msg):
+    cancel_scan()
+    gui.error(msg)
+
 def update(dt=30):
     gui.update(dt)
-    qr_scanner.update()
+    qr_scanner.update(qr_scanner_error)
     usb_host.update()
 
 def ioloop():
--- a/src/qrscanner.py
+++ b/src/qrscanner.py
@@ -190,7 +190,9 @@ class QRScanner:
             QRAnimated.clean()
         try:
             if self.scanning:
-                res = self.uart.read(self.uart.any())
+                res = self.uart.read()
+                if not simulator and res is None:
+                    res = b""
                 if len(res) > 0:
                     self.data += res.decode('utf-8')
                 if self.is_done() and self.callback is not None:
--- a/src/qrscanner.py
+++ b/src/qrscanner.py
@@ -1,5 +1,7 @@
 import utime as time
+import sys
 from platform import simulator
+from io import BytesIO
 
 if not simulator:
     import pyb
@@ -50,6 +52,7 @@ class QRScanner:
         self.scanning = False
         self.t0 = None
         self.callback = None
+        self.callback_animated_qr = None
 
     def query(self, data):
         self.uart.write(data)
@@ -138,12 +141,14 @@ class QRScanner:
         else:
             self.set_setting(SCAN_ADDR, 0)
 
-    def start_scan(self, callback=None, timeout=None):
+    def start_scan(self, callback=None, callback_animated_qr=None, callback_animated_next=None, timeout=None):
         self.trigger_on()
         self.t0 = time.time()
         self.data = ""
         self.scanning = True
         self.callback = callback
+        self.callback_animated_qr = callback_animated_qr
+        self.callback_animated_next = callback_animated_next
 
     def scan(self, timeout=None):
         self.trigger_on()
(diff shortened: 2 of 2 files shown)


stepansnigirev:
Great! I will test it tonight.
We can trigger the scanner again if the data we received is not complete or has a flag that more data is coming (we can use the last few characters for this purpose)
I think the QR scanner we use has a config to ignore the same QR code on the next scan, I need to check.
But relying on this config would break compatibility with other QR scanners.
I think we can still use it but also check if incoming data is the same as before and ignore it if so.

I would suggest to use last 3 characters of the QR code to flag more data. First character - a flag itself, then a byte with the index of the current code, and lastly total number of codes.

As a flag we should use some character that is not in base58, bech32 or base64 sets. Maybe &?

stepansnigirev:
Another thing, if we go with flags we should include this functionality in the qrscanner.py. In this case main logic doesn't need to care about partitioned data - it always gets full data at the end of all scans. Current counter and total counter should be available from the main as we may want to show the user some kind of progress bar.

stepansnigirev:
Sorry, misclicked....

gorazdko:
> Another thing, if we go with flags we should include this functionality in the qrscanner.py. In this case main logic doesn't need to care about partitioned data - it always gets full data at the end of all scans

This is what this PR does - as the #47 describes. The difference is we put the flag into header not in tail:
`"part1of3 payload"`

> I think the QR scanner we use has a config to ignore the same QR code on the next scan

ok, this allows us to skip the button then. I'll check the datasheet...

src/qrscanner.py
```diff
@@ -168,9 +170,33 @@ def update(self):
             if len(res) > 0:
                 self.data += res.decode('utf-8')
             if self.is_done() and self.callback is not None:
+                err_msg = None
                 data = self.data[:-len(self.EOL)]
                 self.reset()
-                self.callback(data)
+                # is this animated QR code
+                if data.startswith("part") and self.callback_animated_qr is not None:
```
gorazdko: here we should probably check if we have 2 words first and then check if the first word starts with "part"
  * this way we certainly avoid collisions so we could also use the shorter version: "p1of3  payload"

stepansnigirev: Makes sense. 
We could use
```py
import re

m = re.match("part\d+of\d+ (.*)",data)
# m.group(0) would be full text if it matches
# m.group(1) will be our payload
# m is None if it didn't have "partMofN"
if m and self.callback_animated_qr:
# ...
```

src/qrscanner.py
```diff
@@ -168,9 +170,33 @@ def update(self):
...
+                            assert indx[1] == QRAnimated.prev_indx[1]
+
+                        self.callback_animated_qr(indx)
+                        QRAnimated.prev_indx = indx
+                        if indx[0] is not indx[1]:
+                            return
+                    except:
+                        # Gui error is thrown in callback
+                        err_msg = "Animated QR: something went wrong!"
+                self.callback(data, err_msg=err_msg)
```
gorazdko: err_msg is causing other callbacks to fail, e.g. Add new wallet. Have to come up with sth different. I wonder if i can use sth like catchit decorator...

stepansnigirev: I was thinking about that... We can either apply `@catchit` to `update()` function in `main`, or if it doesn't work for some reason we can separate `usb_host.update()` and `qr_scanner.update()` to a separate function and use `@catchit` there.

stepansnigirev:
Checked. Works nicely with transaction signing. Breaks other functions like scan new wallet or verify address.
I think if we throw an error from qr scanner and `@catchit` in `update` function instead of sending it to the callback we should be fine. 
Alternatively, we can define constant error callback for qr scanner in the constructor and if an error occurs - call that one instead. It should cancel current scan and show error message.

src/qrscanner.py
gorazdko: I combined stuff from both comments into the code. Just doing regex on a medium sized QR code results in "maximum recursion depth exceeded"

gorazdko:
I still have to test this more tomorrow and will let you know when its ready...

src/qrscanner.py
stepansnigirev: ok, it's another annoying constrain of micropython... Now I remember that I also had this issue some time ago.
Screw regexps, let's just use `data.split(" ")`, check that the first part starts with `part` and split with `of` and get the numbers.

gorazdko: fixed

gorazdko:
> Alternatively, we can set a constant error callback for qr scanner in the constructor and if an error occurs - call that one instead. It should cancel current scan and show error message.

This approach is better as i get to clean the state easier after the exception. However, i had
problems instantiating the object `qr_scanner = QRScanner(cb_error=qr_scanner_error)` this way, so i decided to pass it to update function: `qr_scanner.update(qr_scanner_error)`

Let me know what you think. Otherwise i tested and it seems to work ok

stepansnigirev:
> so i decided to pass it to update function: `qr_scanner.update(qr_scanner_error)`

Sounds good to me. We should know current state of the app and how to handle error properly when we call `qr_scanner.update()`, so it's even more logical than using it in the constructor. `qr_scanner.start()` might also be good for this purpose, but if `update` works - it's all we need for now.

> i couldnt find it in the datasheet. One solution would be for software to check if the new scan is identical to previous scan and ignore and trigger a new scan. For this we would have to disable the beep sound, otherwise it would be bad UX. I would leave this feature for another issue

Yes, let's leave it for later. I think clicking a button for the next scan is not such a big deal, and generally with all the psbt optimizations we did recently number of required QR codes will be pretty small even for huge transactions or wallet descriptors.

What about the simulator? Does it work there as well? You mentioned that there was some problem.

src/qrscanner.py
```diff
@@ -162,15 +166,40 @@ def scan(self, timeout=None):
...
+                    assert QRAnimated.indx[1] == QRAnimated.prev_indx[1], "QRAnim: wrong code"
+                self.callback_animated_qr(QRAnimated.indx)
+                QRAnimated.prev_indx = QRAnimated.indx
+                if QRAnimated.indx[0] is not QRAnimated.indx[1]:
+                    return
+            self.callback(data)
+            QRAnimated.clean()
+        try:
+            if self.scanning:
+                res = self.uart.read(self.uart.any())
```
gorazdko: This line is problematic for the simulator as there are 2 executions of the tcphost.py::_check() : in self.uart.any() and in uart.read()

If i comment out `self.check()` in read() it works ok:

```
    def read(self, nbytes=None):
        #self._check()
        if nbytes is None or nbytes >= len(self.buf):
            buf = self.buf
            self.buf = b''
        else:
            buf = self.buf[:nbytes]
            self.buf = self.buf[nbytes:]
        return buf
```

I can push the change if you're ok with it.

stepansnigirev: I recently discovered that `uart.read()` will return `None` if there is nothing in the buffer, but returns everything available in the buffer otherwise.
Could we use that then?

```py
res = self.uart.read()
if res is None:
    res = b""
```

gorazdko: Fixed

gorazdko:
I have tested on stm32 and unix platform. This is one test example is used on both platforms: 
[test_data.zip](https://github.com/cryptoadvance/specter-diy/files/4321365/test_data.zip)

src/main.py
```diff
@@ -235,13 +235,19 @@ def parse_transaction(b64_tx, success_callback=None, error_callback=None):
...
+    gui.alert("Animated QR: %s/%s" % (indx[0], indx[1]), "Proceed to scanning code No. %s" % str(indx[0]+1), qr_animated_received)
+
 @catchit
 def scan_transaction():
     screens.show_progress("Scan transaction to sign",
                           "Scanning.. Click \"Cancel\" to stop.",
                           callback=cancel_scan)
     gui.update(30)
-    qr_scanner.start_scan(parse_transaction)
+    qr_scanner.start_scan(parse_transaction, qr_animated)
```
stepansnigirev: We can avoid extra callbacks in global scope if we do the following:
```py
def qr_animated(indx, callback):
    gui.alert("Animated QR: %s/%s" % (indx[0], indx[1]), 
              "Proceed to scanning code No. %s" % str(indx[0]+1), 
              callback)

@catchit
def scan_transaction():
    screens.show_progress("Scan transaction to sign",
                          "Scanning.. Click \"Cancel\" to stop.",
                          callback=cancel_scan)
    gui.update(30)
    qr_scanner.start_scan(parse_transaction, qr_animated, scan_transaction)
```

Makes sense to add the same to add new wallet function:
```py
@catchit
def add_new_wallet():    
    screens.show_progress("Scan wallet to add",
                          "Scanning.. Click \"Cancel\" to stop.",
                          callback=cancel_scan)
    gui.update(30)
    qr_scanner.start_scan(parse_new_wallet, qr_animated, add_new_wallet)
```

src/qrscanner.py
```diff
@@ -138,12 +141,13 @@ def trigger_reset(self):
         else:
             self.set_setting(SCAN_ADDR, 0)
 
-    def start_scan(self, callback=None, timeout=None):
+    def start_scan(self, callback=None, callback_animated_qr=None, timeout=None):
```
stepansnigirev: Here we can add another callback parameter then:
```py
def start_scan(self, callback=None, callback_animated_qr=None, callback_animated_next=None, timeout=None):
    # ...
    self.callback_animated_next = callback_animated_next
```

src/qrscanner.py
```diff
@@ -162,15 +166,42 @@ def scan(self, timeout=None):
...
+            # is this animated QR code
+            if qr_anim is not None and self.callback_animated_qr is not None:
+                data = qr_anim
+                assert QRAnimated.indx[0] <= QRAnimated.indx[1], "QRAnim: encoding"
+                if QRAnimated.prev_indx == []:
+                    assert QRAnimated.indx[0] == 1, "QRAnim: wrong first code"
+                else:
+                    assert QRAnimated.indx[0] > QRAnimated.prev_indx[0], "QRAnim: wrong order"
+                    assert QRAnimated.indx[1] == QRAnimated.prev_indx[1], "QRAnim: wrong code"
+                self.callback_animated_qr(QRAnimated.indx)
```
stepansnigirev: and here just pass it to the callback:
```py
self.callback_animated_qr(QRAnimated.indx, self.callback_animated_next)
```

gorazdko:
Force pushed the changes. I also tested signing a transaction and adding a wallet on stm32 and unix platform

stepansnigirev:
Awesome! Thanks! Merged.

PR #55, opened 2020-03-14
gui: animated qr

### LINKED ISSUE #48: Implement displaying animated QR codes
relevant to https://github.com/cryptoadvance/specter-diy/issues/47

Sometimes our QR code is too big. We can split it into pieces and either animate it or ask user to press for the next one.
Should be a class in `gui.components`.

gorazdko:
close #48 

## Status

Ready for review

## Edit

New implementation force pushed which resembles https://github.com/cryptoadvance/specter-desktop/pull/104

### DIFF
--- a/src/gui/components.py
+++ b/src/gui/components.py
@@ -182,10 +182,13 @@ class QRCode(lv.cont):
 	        print("QR on screen:", QrA.header() + self._text)
         self.text_prev = text
 
-        self.lbl.set_size(self.size, self.size)
         if QrA.isQRplaying:
             if self.max_size:
                 self.lbl.set_size(self.max_size, self.max_size)
+            else:
+                self.lbl.set_size(self.size, self.size)
+        else:
+            self.lbl.set_size(self.size, self.size)
 
         qr = qrcode.encode_to_string(QrA.header() + self._text)
         size = int(math.sqrt(len(qr))) # + 4 clear space on every side
--- a/src/gui/components.py
+++ b/src/gui/components.py
@@ -12,6 +12,24 @@ qr_style.text.color = lv.color_hex(0)
 qr_style.text.line_space = 0
 qr_style.text.letter_space = 0
 
+QR_PADDING = const(40)
+
+def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
+    """Helper functions that creates a title-styled label"""
+    if scr is None:
+        scr = lv.scr_act()
+    scr = lv.scr_act()
+
+    if width is None:
+        width = 350
+
+    qr = QRCode(scr)
+    qr.set_text(text)
+    qr.set_size(width)
+    qr.set_text(text)
+    qr.align(scr, lv.ALIGN.IN_TOP_MID, 0, y)
+    return qr
+
 class QRCode(lv.label):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
--- a/src/gui/components.py
+++ b/src/gui/components.py
@@ -6,6 +6,7 @@ import micropython
 from .common import styles
 from math import ceil
 from platform import simulator
+from time import ticks_ms, ticks_diff
 if not simulator:
     import pyb
 
@@ -34,7 +35,6 @@ class QrA:
     """
     Example of animated qr: p1of3 payload
     """
-    timeout = False
     animate_ptr = None
     isQRplaying = False
     isQRtoobig = False
@@ -65,10 +65,7 @@ class QrA:
 
     @staticmethod
     def run():
-        def QrAtimeout(tim):
-            QrA.timeout = True
-        QrA.timer = pyb.Timer(1, freq=3)
-        QrA.timer.callback(QrAtimeout)
+        QrA.timer = ticks_ms()
 
     @staticmethod
     def qr_cb(obj, event):
@@ -82,13 +79,11 @@ class QrA:
     def handle():
         # call in main loop
         if QrA.abort:
-            if QrA.timer:
-                QrA.timer.deinit()
-                QrA.timeout = False
+            QrA.timer = None
             QrA.isQRplaying = False
             QrA.abort = False
-        elif QrA.timeout == True:
-            QrA.timeout = False
+        elif QrA.timer != None and ticks_diff(ticks_ms(), QrA.timer) >= 300:
+            QrA.timer = ticks_ms()
             QrA.animate_ptr()
 
     @staticmethod
--- a/src/gui/components.py
+++ b/src/gui/components.py
@@ -2,6 +2,12 @@ import lvgl as lv
 import qrcode
 import math
 import gc
+import micropython
+from .common import styles
+from math import ceil
+from platform import simulator
+if not simulator:
+    import pyb
 
 qr_style = lv.style_t()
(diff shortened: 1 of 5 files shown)


stepansnigirev:
I think it would be better to extend QRCode class defined in
https://github.com/cryptoadvance/specter-diy/blob/master/src/gui/components.py#L15

We could add a method to set max size, and if QR code text is larger than that size - add two arrows on the right and left sides of the qr code. When you click on the arrow - next / prev part of the QR code is shown.
Label under the QR code could show current and total number of qr codes.

What do you think?

stepansnigirev:
Maybe something like this:

Screenshot 2020-03-14 at 14 22 12

src/gui/components.py
```diff
@@ -26,23 +29,98 @@ def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
     qr = QRCode(scr)
     qr.set_text(text)
     qr.set_size(width)
-    qr.set_text(text)
```
gorazdko: What is the purpose of this hack? I see that QR gets squeezed a little. But i dont understand why it has to be called again when it is called inside qr.set_size()...

src/gui/components.py
```diff
@@ -26,23 +29,98 @@ def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
...
+    qr.set_text(None)
     qr.align(scr, lv.ALIGN.IN_TOP_MID, 0, y)
+
+    if qr.buttons != None:
+        # FIXME dont know where the QR code is
+        assert qr.indx_msg != None
+        BTN_PADDING = 30
+        qr.buttons[0].align(scr, lv.ALIGN.IN_TOP_MID, -(width//2 + BTN_PADDING), width//2)
+        qr.buttons[1].align(scr, lv.ALIGN.IN_TOP_MID, (width//2 + BTN_PADDING), width//2)
+        qr.indx_msg.align(qr, lv.ALIGN.IN_TOP_MID, 0, 20)
```
gorazdko: i will have to throw this out as i dont know where the qr is to align it with the buttons. The buttons can only be aligned in QRAlert(). So add_qrcode() will always have to be called in QRAlert. Yes?

src/gui/components.py
```diff
@@ -26,23 +29,98 @@ def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
...
     
-    def set_text(self, text="Text"):
-        if self._text != text:
-            print("QR on screen:", text)
-        self._text = text
-        qr = qrcode.encode_to_string(text)
+    def set_text(self, text=None):
+        if text != None and QRCode.is_qr_too_big(text):
+            scr = lv.scr_act()
+            self.buttons = add_button_pair(text1="<", text2=">", scr=scr,
```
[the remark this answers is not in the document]
gorazdko: i did try that but then the button gets chopped off.

Screenshot from 2020-03-18 15-21-37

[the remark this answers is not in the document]
stepansnigirev: Ah, I see. Maybe we can wrap everything in a container then.
At the moment QRCode inherits from the `lv.label`,

Let's make QRCode inherit from a `lv.obj` (our container), inside we would have:
- a qr_label that is a QR code itself (or part of it)
- two buttons with lv.SYMBOL.LEFT and lv.SYMBOL.RIGHT as labels
- index label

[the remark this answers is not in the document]
gorazdko: fixed

src/gui/components.py
```diff
@@ -28,23 +31,116 @@ def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
...
+            return self.indx != []
+        return text in self.payload or text == "".join(self.payload)
+
+    def clean(self):
+        self.indx = []
+        self.payload = []
+
+class QRCode(lv.cont):
+    # number of bytes to be QR encoded
+    MAX_SIZE = const(300) # FIXME
```
gorazdko: This should be large enough so that we dont unnecessarily animate other stuff, e.g. MASTER SINGLE KEY animation will cause overlapping of the message with the show-derivation-path slider. Also, small animations look a bit different (tight) due to a bit unpredictable QR padding algorithm

gorazdko:
I force pushed the implementation which now works the same as https://github.com/cryptoadvance/specter-desktop/pull/104

What should we do with animations in simulator mode? It doesnt look to have a timer.

src/gui/components.py
```diff
@@ -18,34 +24,178 @@ def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
...
+        self.cont.set_fit(lv.FIT.TIGHT)
+        self.set_fit(lv.FIT.TIGHT)
+        self.isQRbig = False
+        self.animation_tip = True
+
+        QrA.init(self.next)
+        self.set_event_cb(QrA.qr_cb)
+
+    def is_qr_big(self, message):
+        QrA.isQRtoobig = len(message) >= 850
```
gorazdko, line 145: I set the limit at 850 which may be even lower. It needs more testing. The thing is drawing the qr takes alot of RAM, more than 10 times of the qr payload. One thing i noticed also, we can draw a 1KB qrcode in a 300x300 label most of the times but not into a 480x480 label because it requires that much more RAM.

stepansnigirev: I think to solve it we could move lvgl buffer to SDRAM in the display driver.
I opened a corresponding issue: https://github.com/diybitcoinhardware/f469-disco/issues/8

For now let's just live with this limitation.

stepansnigirev:
> What should we do with animations in simulator mode? It doesnt look to have a timer.

We can write a simple scheduler where we can register / unregister callbacks. Maybe even look at `asyncio`.
I think we can do the same both in simulator and on the board itself.
At the moment we have something very similar to asyncio in `ioloop()` function - we constantly call `.update()` function on everything that requires updates.

Here is a pure python implementation:
https://github.com/micropython/micropython-lib/blob/master/asyncio_slow/asyncio_slow.py

Recently `uasyncio` was added to micropython:
https://github.com/micropython/micropython/pull/5332

We can also rebase to the latest version and check that nothing breaks.

gorazdko:
> Maybe even look at asyncio

For now I just used ticks.diff for simulator and for stm32 with the .update function, so the code is the same. Is that ok?

In simulator i notice this qr width glitching, it doesnt happen on hardware though:

qrA

Its a consequence of the fact that i have to adapt the qr width after i draw it. Will ponder on this one a bit...

stepansnigirev:
Looks good! I will test a bit a merge :)

stepansnigirev:
Tested works great, I didn't see any width glitches in the simulator.
On hardware the close button becomes a bit laggy but I think it's ok, we can live with it.

With https://github.com/cryptoadvance/specter-desktop/pull/118 we can use it in specter-desktop, and now we can sign and display large transactions!

gorazdko:
> On hardware the close button becomes a bit laggy but I think it's ok, we can live with it

You mean pressing OK button while animation is running? Weird, I havent noticed that at all. Mine has never lagged and works blazingly fast.

k9ert:
This should also go into the release, i guess? I'd suggest to quickly merge it and then test everything together?!

stepansnigirev:
I agree, works great, I don't see a reason not to merge :)

PR #58, opened 2020-04-14
qrscanner: scan qr animations automatically.

### LINKED ISSUE #56: qr code animations
## Abstract

I did a proof of concept for scanning animated qrs automatically (without buttons): https://github.com/gorazdko/specter-diy/tree/animiated_qr_video
  *  scanner is scanning in continuous mode with the following parameters:
```
# 100ms
INTERVAL_OF_SCANNING = 0x01
# 0x32 - enable 5 second delay
DELAY_OF_SAME_BARCODES = 128 + 0x32
```
Animations can be played with https://github.com/gorazdko/specter-desktop/tree/qr_animation
  *  `def qr_animate(qrtext, max_len=100):` max_len determines the maximum chunk size of the
frame
  *  interval of animating can be set in `window.setInterval(function(){full_psbt.qra_next()}, 250)`
  * qrs are set for high recovery level (30%): `QRCode.CorrectLevel.H`

## Results

Its feasible, though results arent nearly as good as in https://divan.dev/posts/animatedqr/ where 

> The best result was 1.4 secs, which is almost 9KB/s! This result has been recorded at a rate of 11 frames per second and chunk size of 850 bytes with Medium recovery level

Here, one setting that works ok is 250 ms interval with 100 byte frames. For optimal settings one should play with parameters

Should we prefer this method over the one with buttons?

gorazdko:
## Abstract

Close #56 

After correcting some bugs, the scanning is now faster and more fluent.

## Status

* ~ready for review~
* does not support the *expensive* scanner
* ~wip: add pin trigger mode~
* ready for review

### DIFF
--- a/src/qrscanner.py
+++ b/src/qrscanner.py
@@ -2,6 +2,7 @@ import utime as time
 import sys
 from platform import simulator
 from io import BytesIO
+from ubinascii import hexlify
 
 if not simulator:
     import pyb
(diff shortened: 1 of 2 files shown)


stepansnigirev:
Looks great! Let me test a bit more over the next few days and I will merge :)

We can drop support for an "expensive" scanner as nobody uses it anyway. Let's keep only minimal support for other scanners (no animations if qr scanner is in pin trigger mode).
For these scanners there is always a possibility to scan non-animated QR code, and we should keep this as an option in specter-desktop - just because there are also other wallets that don't use animated QR codes.

I really like the feeling of scanning animated QR codes! Huge improvement in the UX.

src/qrscanner.py
```diff
@@ -15,42 +16,56 @@
...
-        # we autoconfigure scanner before the first scan, 
-        # as we don't know if scanner already got power or not
-        # during the constructor
-        # and if we have scanner at all
-        if trigger is not None or simulator:
-            self.trigger = pyb.Pin(trigger, pyb.Pin.OUT)
-            self.trigger.on()
-            self.is_configured = True
-
+        self.init()
```
gorazdko, line 69: i havent noticed any problems with initializing here in constructor. If you know there are problems i rather call this as a callback after 2 seconds

gorazdko:
I forced pushed changes. I tested pin triggered mode only with my scanner which is tricky. E.g. if you flash the code with pin triggered mode, it will not work as the baudrate in scanner is still 115200 and we're operating with 9600.

But i dont know if this works with different scanners.

stepansnigirev:
Great, it works like a charm!
I was able to import a wallet with 10 cosigners and successfully signed a transaction with it.
QR code scanning works very well.

Merging!

PR #59, opened 2020-05-16
check for address gap limit violation

### LINKED ISSUE #50: Keep track of receiving and change addresses
Software wallets often have a gap limit of 20 addresses, so if my last change address was `m/.../1/123` and in the next transaction I use a change `m/.../1/9999` it can cause lock of funds.
We should **warn** the user if the change index is weird. To do this we should store the largest change index and warn if the next one is more than 20 larger.

gorazdko:
## Abstract

Close #50 

## Status

~WIP~ Ready for review:
- [x] final testing
- [x] extracting indexes from compressed and full PSBTs
- [x] noticed also: on verify address screen cancel button does not work if running telnet
   * opened a new issue #60
- [x] when creating a new wallet initialize indexes with -1, so that we correctly browse wallet addresses. If we initialize with 0  we cant know if the address was used already or not
- [x] when browsing wallet addresses we currently allow to go back to address Nr. 1 regardless of max indexes . Should we disallow that or warn that these addresses may have been used already? -> Warnings added.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -181,6 +181,7 @@ def xpubs_menu():
 def sign_psbt(wallet=None, tx=None, success_callback=None):
     wallet.fill_psbt(tx)
     keystore.sign(tx)
+    keystore.update_wallet_indexes(wallet, tx)
     # remove everything but partial sigs
     # to reduce QR code size
     tx.unknown = {}
--- a/src/main.py
+++ b/src/main.py
@@ -300,6 +300,9 @@ def verify_address(s):
             if index > w.last_rcv_idx + w.gap_limit:
                 warning = ("\n\n #ff0000 Possible gap limit. Using this address# "
                            "#ff0000 may lead to locking of your funds!#")
+            elif index <= w.last_rcv_idx:
+                warning = ("\n\n #ff0000 This address may have been used before.#\n"
+                           "#ff0000 Reusing it would diminish your privacy!#")
             popups.qr_alert("Address #%d from wallet\n\"%s\"" % (index+1, w.name),
                             "bitcoin:%s"%addr, message_text=addr + warning)
             return
--- a/src/main.py
+++ b/src/main.py
@@ -296,8 +296,12 @@ def verify_address(s):
         return
     for w in keystore.wallets:
         if w.address(index) == addr:
+            warning = ""
+            if index > w.last_rcv_idx + w.gap_limit:
+                warning = ("\n\n #ff0000 Possible gap limit. Using this address# "
+                           "#ff0000 may lead to locking of your funds!#")
             popups.qr_alert("Address #%d from wallet\n\"%s\"" % (index+1, w.name),
-                            "bitcoin:%s"%addr, message_text=addr)
+                            "bitcoin:%s"%addr, message_text=addr + warning)
             return
     gui.error("Address doesn't belong to any wallet. Wrong device or network?")
 
--- a/src/main.py
+++ b/src/main.py
@@ -231,9 +231,33 @@ def parse_transaction(b64_tx, success_callback=None, error_callback=None):
         if error_callback is not None:
             error_callback("invalid argument")
         return
+
+    # check for address gap limit
+    gap_limit = False
+    w = data["wallet"]
+    wallet_key = b"\xfc\xca\x01" + data["wallet"].fingerprint
+    for el in tx.inputs + tx.outputs:
+        if el.bip32_derivations:
+            # full PSBT
+            for der in el.bip32_derivations:
+                drv = el.bip32_derivations[der].derivation
+                if drv[-1] > w.gap_limit + (w.last_chg_idx if drv[-2] else w.last_rcv_idx):
+                    gap_limit = True
+        elif wallet_key in el.unknown:
+            # compressed PSBT
+            idxs = el.unknown[wallet_key]
+            chg = int.from_bytes(idxs[0:4], 'little')
+            idx = int.from_bytes(idxs[4:8], 'little')
+            if idx > w.gap_limit + (w.last_chg_idx if chg else w.last_rcv_idx):
+                gap_limit = True
+    if gap_limit:
+        data["warning"] = ("#ff0000 Possible gap limit exceeded with some#\n"
+                           "#ff0000 addresses. Your funds may get locked. #\n "
+                           "#ff0000 Proceed at your own risk!#")
+
     title = "Spending %u\nfrom %s" % (data["spending"], data["wallet"].name)
     popups.prompt_tx(title, data,
-        ok=cb_with_args(sign_psbt, wallet=data["wallet"], tx=tx, success_callback=success_callback), 
+        ok=cb_with_args(sign_psbt, wallet=data["wallet"], tx=tx, success_callback=success_callback),
         cancel=cb_with_args(error_callback, "user cancel")
     )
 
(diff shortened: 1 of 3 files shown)


stepansnigirev:
Looks good! Gonna test tonight.

> when browsing wallet addresses we currently allow to go back to address Nr. 1 regardless of max indexes . Should we disallow that or warn that these addresses may have been reused already?

I think we should warn that the address might be reused.

src/main.py
```diff
@@ -237,8 +237,8 @@ def parse_transaction(b64_tx, success_callback=None, error_callback=None):
         for el in scope:
             for der in el.bip32_derivations:
```
gorazdko, line 238: weird. On stm32 i cant get into this loop because el.bip32_derivations is empty. Works in simulator though.
I copied the lines from you:
 ```
    # blue wallet trick - if the fingerprint is 0 we use our fingerprint
    for scope in [tx.inputs, tx.outputs]:
        for el in scope:
            for der in el.bip32_derivations:
                if el.bip32_derivations[der].fingerprint == b'\x00\x00\x00\x00':
                    el.bip32_derivations[der].fingerprint = keystore.fingerprint
```
and i wanted to ask you if you tested it on hw, but it seems this code doesnt do anything, its dead.

Comment (stepansnigirev) [reply src/main.py:238]: It's not dead. It works only if fingerprint in bip32 derivation of PSBT is zero - Bitcoin Core will never produce such transaction, but Blue Wallet doesn't know our fingerprint so they just fill zeroes there. If you try with Blue Wallet or manually set fingerprint to \x00\x00\x00\x00 you will see that this code is actually replacing zero fingerprint with the fingerprint of the device.

Comment (stepansnigirev) [reply src/main.py:238]: And yeah, I tested that tx signing with blue wallet works, so it should enter this loop.
bip39 derivation is empty probably because we delete all unnecessary fields after signing.

Comment (stepansnigirev) [reply src/main.py:238]: Here we delete bip39 derivation. 
https://github.com/cryptoadvance/specter-diy/pull/59/files#diff-0c81b9c94c68ea34c497c1fa52daee56R194
So you need to do index update before this happens

Comment (gorazdko) [reply src/main.py:238]: what am i missing then. Who/what is using variable *el*?

Comment (stepansnigirev) [reply src/main.py:238]: It fills the fingerprint in the original transaction. 
`el` is an instalnce of PSBTInput/Output in the list, `bip32_derivation` is a dictionary.
So by when class instance is copied to the variable `el`,  `bip32_derivation` is not a copy of origianl dict, but a reference. 
So by changing `el.bip32_derivations[der].fingerprint` we are actually changing corresponding part in the original transaction: `tx.inputs[i].bip32_derivations[der].fingerprint`.

I agree that this piece of code is very confusing.
It might be more understandable if we change input and output directly using index with `enumerate` and maybe use two identical loops through inputs and outputs.

Comment (gorazdko) [reply src/main.py:238]: Amazing.

> so it should enter this loop.

For me it doesnt enter that loop. This code asserts only in simulator, but not on hardware. I scan exactly the same transaction on both platforms and both platforms contain the same userdata folder. Anyways, ill debug tomorrow and try to resolve this...

```
    # blue wallet trick - if the fingerprint is 0 we use our fingerprint
    for scope in [tx.inputs, tx.outputs]:
        for el in scope:
            print ("+++++++++++++")
            for der in el.bip32_derivations:
                assert(0)
                if el.bip32_derivations[der].fingerprint == b'\x00\x00\x00\x00':
                    el.bip32_derivations[der].fingerprint = keystore.fingerprint
```

gorazdko: I figured out the problem: 

scr

I thought the payloads in QRcode and in text were supposed to be the same.

COPIED:

`cHNidP8BAIkCAAAAAdGUzLtZzKFFaKwlJ/6i9SoClwIl9D39o4jJ0GGREHypAAAAAAD+////AlQKDY8AAAAAIgAgZECaMwzPqUrlqKQV1f13LR4tJB4wRm6k8fJJMw5jBPAA4fUFAAAAACIAIDDbsdJT8/jJ9Hc9h9gkaQxESiRM4H12HT13roHnvWITAAAAAE8BBDWHzwQixZjegAAAAiHVE6AJMxGZaJzTEtGe8d52W3DDHq1m2GWcxebVQeYbApwt0keBnXaiR+HPD9AGmGu44S7DXpwG6G1irgMkJw++FJ0dxgSAAAAwgAAAAYAAAACAAAACTwEENYfPBIwihT6AAAACEttNnBEfJ/EDUr86z1GCgYFvFzSgm0bbYg93s72I4OsCxDvJhefZEcPqBtFqvvQr4IYV7brPRy+hU8nzX35QULwU+HIPjoAAADCAAAABgAAAAIAAAAIAAQErAPkClQAAAAAiACBnx0f/wIzvwdFA7bRBfNE50U/dhy7RY7umbJIBHdz8owEFR1EhA3syvd5QtweU1o2w6vZmgcUT1/+Qi2s771bzIVmz2+kuIQPh4V5YF2smwV+3l+UolCro0t/ZgsHgS5mWwikzsp5k+FKuIgYDezK93lC3B5TWjbDq9maBxRPX/5CLazvvVvMhWbPb6S4cnR3GBDAAAIABAACAAAAAgAIAAIAAAAAAAAAAACIGA+HhXlgXaybBX7eX5SiUKujS39mCweBLmZbCKTOynmT4HPhyD44wAACAAQAAgAAAAIACAACAAAAAAAAAAAAAAQFHUSECAQbB7bMRODRTRLfxIZ/eFqaOQYF1/rV51OGifKmEj7YhA9qgAEktNr5tUR0CKTxDkKXhkEuFwr+f4PbjaOUj3JywUq4iAgIBBsHtsxE4NFNEt/Ehn94Wpo5BgXX+tXnU4aJ8qYSPthz4cg+OMAAAgAEAAIAAAACAAgAAgAEAAAAAAAAAIgID2qAASS02vm1RHQIpPEOQpeGQS4XCv5/g9uNo5SPcnLAcnR3GBDAAAIABAACAAAAAgAIAAIABAAAAAAAAAAABAUdRIQJAHnTn9EHrzwQSrJL4Kzw0KY+vNYRcXUZdL04stGLAPyEDEF1dAqlVsS5FHisIs+YbaR0BWwNhWjQRaQl7fEIlMhxSriICAkAedOf0QevPBBKskvgrPDQpj681hFxdRl0vTiy0YsA/HPhyD44wAACAAQAAgAAAAIACAACAAAAAAAwAAAAiAgMQXV0CqVWxLkUeKwiz5htpHQFbA2FaNBFpCXt8QiUyHBydHcYEMAAAgAEAAIAAAACAAgAAgAAAAAAMAAAAAA==`


SCANNED:

`cHNidP8BAIkCAAAAAdGUzLtZzKFFaKwlJ/6i9SoClwIl9D39o4jJ0GGREHypAAAAAAD+////AlQKDY8AAAAAIgAgZECaMwzPqUrlqKQV1f13LR4tJB4wRm6k8fJJMw5jBPAA4fUFAAAAACIAIDDbsdJT8/jJ9Hc9h9gkaQxESiRM4H12HT13roHnvWITAAAAAAABASsA+QKVAAAAACIAIGfHR//AjO/B0UDttEF80TnRT92HLtFju6ZskgEd3PyjB/zKASOiEV8IAAAAAAAAAAAAB/zKASOiEV8IAQAAAAAAAAAAB/zKASOiEV8IAAAAAAwAAAAA`

stepansnigirev: There is a special custom field in compressed psbt that contains derivation index.
It is stored in "unknown" dict, the key value is b"\xfc\xca\x01"+wallet.fingerprint, derivation path is two 4-byte indexes - change and address index.

Here is an example how I process it:
https://github.com/cryptoadvance/specter-diy/blob/master/src/keystore.py#L330

gorazdko:
Fixed and force-pushed changes.

stepansnigirev:
I tested, works great! Merging!

PR #83, opened 2020-09-24
Add SDCard and Smartcard support for key storage

stepansnigirev:
Flexible security model that you can set up at first boot:

# Approach 1
Don't store recovery phrase (amnesic mode, forgets key when turned off), store it on internal flash (Trezor security model) or encrypted on a separate SD card (only this device will be able to decrypt the file)

# Approach 2
Amnesic or store the key on a PIN-protected javacard. Uses [`MemoryCardApplet`](https://github.com/cryptoadvance/specter-javacard/blob/master/docs/MemoryCard.md) for that, requires extension board with the smartcard slot.
Javacards have reasonably good hardware security - they are tamper-resistant, detect all kind of glitches and introduce additional countermeasures against sidechannel attacks. But they are running proprietary JavaCardOS.
We still can benefit from their security features.

# TODO:

- [x] add support for Javacards including secure communication.
- [x] encrypt Bitcoin secret such that it's not stored on the Javacard directly - then the attacker will need to hack both Specter-DIY and Javacard.
- [x] add support for SD card as a secret storage
- [x] merge SDKeyStore and FlashKeyStore to a single class and allow the user to chose what media to save the key to.
- [x] add wipe functionality that effectively works as factory reset
- [x] fix plug-unplug of the smartcard (should reopen channel and prompt for PIN code instead of crashing)


bitcoinheiro:
This is impressive, great work!

PR #87, opened 2020-10-19
correct name and add link to 3d printed box for storage

bitcoinheiro:
Keep it all exposed, but protect from dust and other environment degradation


bitcoinheiro:
Also, FYI - I've made this Specter logo for 3d printing...
https://www.tinkercad.com/things/7oI3nDeyorQ

You can download it and use it to stamp any specter related 3d print

stepansnigirev:
This is so awesome! And sorry for a mistake in your name...

PR #92, opened 2020-10-29
Add specify BIP44 account option to export screen

ben-kaufman:
The Master public keys screen already allows some popular default paths and custom derivations, so this PR is just a small convenience suggestion. I added a button there to specify the BIP44 account which will be used to derive the default xpubs, so the user don't need to do it manually for every account with the custom derivation option.

Screen Shot 2020-10-29 at 10 07 57

Screen Shot 2020-10-29 at 10 08 06

### DIFF
--- a/src/apps/xpubs/xpubs.py
+++ b/src/apps/xpubs/xpubs.py
@@ -1,5 +1,5 @@
 from app import BaseApp, AppError
-from gui.screens import Menu, DerivationScreen
+from gui.screens import Menu, DerivationScreen, NumericScreen
 from .screens import XPubScreen
 
 from binascii import hexlify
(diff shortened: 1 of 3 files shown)


src/apps/xpubs/xpubs.py
```diff
@@ -19,35 +19,41 @@ class XpubApp(BaseApp):
...
+                    "m/48h/%dh/%dh/1h" % (coin, self.account),
+                    "Multisig Nested Segwit\nm/48h/%dh/%dh/1h" % (coin, self.account),
                 ),
             ]
         else:
-            buttons += [(0, "Show more keys"), (1, "Enter custom derivation")]
+            buttons += [(0, "Show more keys"), (2, "Change account number"), (1, "Enter custom derivation")]
         # wait for menu selection
-        menuitem = await show_screen(Menu(buttons, last=(255, None)))
+        menuitem = await show_screen(Menu(buttons, last=(255, None), title="Select the key"))
```
stepansnigirev, line 56: Let's add the current account number to the menu note as well:
```suggestion
        menuitem = await show_screen(Menu(buttons, last=(255, None),
                                          title="Select the key",
                                          note="Current account number: %d" % self.account))
```

PR #137, opened 2021-03-30
Sorting files

xavierfiechter:
1.txt, 2.txt, 3.txt instead of 
2.txt, 3.txt, 1.txt ... 

It's not natsort, but better than a random? representation.


stepansnigirev:
Makes sense. Thanks!

PR #143, opened 2021-04-28
add Amazon pins

mflaxman:
If you're already buying the board from mouser it makes sense to get the pins from them, but if for people in the US who need long pins this is an Amazon prime option.

For example, I didn't buy pins because I already had some, but they were regular length which was too short. These Amazon pins worked well enough.


stepansnigirev:
Thanks!

PR #146, opened 2021-05-07
Added cancel button for pin change

EndNationStates:
Added a cancel button for 'Change device PIN' screen. 
Moved the 'confirm' button in initial PIN screen to bottom of screen for consistency.
Disabled submit when no value is entered to prevent unnecessary failed PIN attempts.

Screenshot 2021-05-08 at 09 06 23

Screenshot 2021-05-08 at 09 13 13

Edit: Fixed element positions

### DIFF
--- a/src/gui/screens/input.py
+++ b/src/gui/screens/input.py
@@ -258,7 +258,7 @@ class PinScreen(Screen):
     def cb(self, obj, event):
         if event == lv.EVENT.RELEASED:
             c = obj.get_active_btn_text()
-            if c is None:
+            if c is None or c == " ":
                 return
             if c == lv.SYMBOL.CLOSE:
                 self.reset()
--- a/src/gui/screens/input.py
+++ b/src/gui/screens/input.py
@@ -185,14 +185,14 @@ class PinScreen(Screen):
         if subtitle is not None:
             lbl = add_label(subtitle, scr=self, style="hint")
             lbl.set_recolor(True)
-            lbl.align(self.title, lv.ALIGN.OUT_BOTTOM_MID, 0, 0)
+            lbl.align(self.title, lv.ALIGN.OUT_BOTTOM_MID, 0, 10)
         if note is not None:
             lbl = add_label(note, scr=self, style="hint")
-            lbl.align(self.title, lv.ALIGN.OUT_BOTTOM_MID, 0, 180)
+            lbl.align(self.title, lv.ALIGN.OUT_BOTTOM_MID, 0, 110)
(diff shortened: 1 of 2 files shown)


src/gui/screens/input.py
```diff
@@ -260,6 +273,15 @@ def cb(self, obj, event):
...
 
+    def submit(self):
+        if self.get_value() == "":
+            return
+        self.release()
+
+    def cancel(self):
+        # obj.del_async()
+        self.pin.set_text(self.CANCEL_VALUE)
+        self.set_value(None)
```
stepansnigirev, line 284:
```suggestion
        self.release()
```

src/gui/screens/input.py
```diff
@@ -260,6 +273,15 @@ def cb(self, obj, event):
     def get_value(self):
         return self.pin.get_text()
```
stepansnigirev, line 274: I think it's more logical to return `None` instead of some custom CANCEL_VALUE here:
```suggestion
        v = self.pin.get_text()
        return None if v == self.CANCEL_VALUE else v
```

`CANCEL_VALUE` is still used internally in the PIN screen, but at least other components don't need to know about it.

src/keystore/ram.py
```diff
@@ -328,15 +330,18 @@ async def setup_pin(self, get_word=None):
 
     async def change_pin(self):
         # get_auth_word function can generate words from part of the PIN
-        old_pin = await self.get_pin(title="First enter your old PIN code")
+        old_pin = await self.get_pin(title="First enter your old PIN code", with_cancel=True)
+        if old_pin == PinScreen.CANCEL_VALUE:
```
stepansnigirev, line 334: With the change above in `get_value` method (returning None instead of custom value) we can check against `None` here:
```suggestion
        if old_pin is None:
```

stepansnigirev (commented):
Thanks, looks good! I left a few minor comments.

src/keystore/ram.py
EndNationStates: Tried that already: TypeError: can't convert 'NoneType' object to str implicitly

src/gui/screens/input.py
```diff
@@ -260,6 +273,15 @@ def cb(self, obj, event):
...
 
+    def submit(self):
+        if self.get_value() == "":
+            return
+        self.release()
```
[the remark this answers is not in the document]
EndNationStates: Reverted.

EndNationStates:
Thanks for the feedback!

Issue #44, opened 2020-02-22
SD card support


stepansnigirev:
We have SD card slot, we can use it.

What we would need:
- check if SD card is there
- mount SD card, read file from it
- save file to SD card
- unmount
- repeat

At the moment if you turn on the device with SD card, Micropython will try to load `main.py` file from SD card and execute it. It can be solved by placing a special file to the flash - don't remember the name exactly, something like NOSD.

If this file is in flash SD card will be ignored on boot and firmware will be loaded from flash. 
After that I **can** do the steps above **once**, but as soon as I unmount the SD card I can'd mount it again anymore.

I think this requires changing micropython itself, in particular, our fork: https://github.com/diybitcoinhardware/micropython

stepansnigirev:
Partially solved with #83 , for now only for key storage.
But looks like all the problems are solved so we could load PSBTs from SD card as well.

SaroC:
Issue:
- Initial FW initial_firmware_v1.8.3.bin flashed with `st-flash write ~/development/specter-bins/initial_firmware_v1.8.3.bin 0x8000000`
- If I boot the device simply (without a SD card inserted into the socket) all works properly I insert the PIN and the OS show me the menu.
- If I boot the device with a SD card into the socket (FAT formatted)...After that I insert my PIN I receive the follow output:
  `Error! Somenthing unexpected happened...
  Traceback (most recent call last):
  File "specter.py", line 76, in handle_exception
  File "specter.py", line 156, in main
  File "specter.py", line 186, in initmenu
  File "keystore/sdcard.py", line 110, inis_key_saved
  File "platform.py", line 95, in mount_sdcard
  OSError: [Errno 19] ENODEV`

The 32GB SD card was formatted with Win (FAT32 with 16kB allocating size) first and also with Linux's mkfs program (with `-t vfat argument`). The SD card was used with and without file inside in.
If I remove the SD card, also while the device is in on state, the FW works properly again showing me the menu again.
Seems that the presence of the SD card is enough to cause the error.

 Any suggestion? maybe the size of SD card could be the root cause?  a smaller one could fix the issue?
 Many Thanks for the support.

schnuartz-ai:
## Audit of current HEAD (3f3c831)

I investigated this tracking issue against the current code. Summary of the original TODO list vs. current implementation:

| TODO item | Status | Evidence |
| --- | --- | --- |
[...]

PR #148, opened 2021-05-21
export wallet

EndNationStates:
Added wallet export in plain text and CC supported json #44

### DIFF
--- a/src/apps/xpubs/xpubs.py
+++ b/src/apps/xpubs/xpubs.py
@@ -56,7 +56,7 @@ class XpubApp(BaseApp):
             ]
         else:
             buttons += [(0, "Show more keys"), (2, "Change account number"), (1, "Enter custom derivation"),
-                        (3, "Export all keys")]
+                        (3, "Export all keys to SD")]
         # wait for menu selection
         menuitem = await show_screen(Menu(buttons, last=(255, None),
                                           title="Select the key",
--- a/src/apps/xpubs/xpubs.py
+++ b/src/apps/xpubs/xpubs.py
@@ -93,7 +93,7 @@ class XpubApp(BaseApp):
                     filedata = ""
                     for der in derivations:
                         xpub = self.keystore.get_xpub(der[2])
-                        filedata += "[%s/%s]%s\n" % (fingerprint, der[2], xpub.to_base58(net["xpub"]))
+                        filedata += "[%s/%s]%s\n" % (fingerprint, der[2][2:].replace("'","h"), xpub.to_base58(net["xpub"]))
                     extension = "txt"
                 else:
                     m = self.keystore.get_xpub("m")
--- a/src/apps/xpubs/xpubs.py
+++ b/src/apps/xpubs/xpubs.py
@@ -196,7 +196,7 @@ class XpubApp(BaseApp):
         platform.unmount_sdcard()
 
     async def save_menu(self, show_screen):
-        buttons = [(0, "Specter-DIY"), (1, "Cold Card")]
+        buttons = [(0, "Specter-DIY (plaintext)"), (1, "Cold Card (json)")]
         # wait for menu selection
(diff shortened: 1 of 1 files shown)


src/apps/xpubs/xpubs.py
```diff
@@ -69,10 +73,59 @@ async def menu(self, show_screen, show_all=False):
...
+                        "account": self.account,
+                        "chain": "BTC" if self.network == "main" else "XTN"
+                    }
+
+                    for der in derivations:
+                        data[der[0]] = {
+                            "name": der[1],
+                            "deriv": der[2],
+                            "xpub": self.calculate_xpub(der[2]),
+                            "_pub": self.calculate_xpub(der[2], True),
```
stepansnigirev, line 111: Calculating xpubs is an expensive function (it does some EC math in `keystore.get_xpub`),
so better to call `keystore.get_xpub` once and then serialize it to xpub and slip132.
I'd suggest deriving the key in this function once per derivation path, and then serializing it with proper version.

EndNationStates: Will rewrite

src/apps/xpubs/xpubs.py
```diff
@@ -123,19 +179,38 @@ async def show_xpub(self, derivation, show_screen):
...
+    def write_file(self, filename, filedata):
+        if not platform.is_sd_present():
+            raise AppError("SD card is not present")
+        platform.mount_sdcard()
+        with open(platform.fpath("/sd/%s" % filename), "w") as f:
+            f.write(filedata)
+        platform.unmount_sdcard()
+
+    async def save_menu(self, show_screen):
+        buttons = [(0, "Specter-DIY"), (1, "Cold Card")]
```
stepansnigirev, line 199:
```suggestion
        buttons = [(0, "Specter-DIY (plaintext)"), (1, "Cold Card (json)")]
```

src/apps/xpubs/xpubs.py
```diff
@@ -69,10 +73,62 @@ async def menu(self, show_screen, show_all=False):
...
+                    ('bip44', "p2pkh", "m/44'/%d'/%d'" % (coin, self.account)),
+                    ('bip48_2', "p2wsh", "m/48'/%d'/%d'/2'" % (coin, self.account)),
+                    ('bip48_1', "p2sh-p2wsh", "m/48'/%d'/%d'/1'" % (coin, self.account)),
+                ]
+
+                if format == self.export_specter_diy:
+                    filedata = ""
+                    for der in derivations:
+                        xpub = self.keystore.get_xpub(der[2])
+                        filedata += "[%s/%s]%s\n" % (fingerprint, der[2], xpub.to_base58(net["xpub"]))
```
stepansnigirev, line 96: Right now it writes `[fingerprint/m/deriavtion]xpub`, shoudl be `[fingerprint/derivation]xpub`. Also we noramlly use `h` for hardened derivation. We should use `'` only for compatibility with coldcard in json files.
```suggestion
                        filedata += "[%s/%s]%s\n" % (fingerprint, der[2][2:].replace("'","h"), xpub.to_base58(net["xpub"]))
```

src/apps/xpubs/xpubs.py
```diff
@@ -52,7 +55,8 @@ async def menu(self, show_screen, show_all=False):
                 ),
             ]
         else:
-            buttons += [(0, "Show more keys"), (2, "Change account number"), (1, "Enter custom derivation")]
+            buttons += [(0, "Show more keys"), (2, "Change account number"), (1, "Enter custom derivation"),
+                        (3, "Export all keys")]
```
stepansnigirev, line 59:
```suggestion
                        (3, "Export all keys to SD")]
```

stepansnigirev (commented):
Last few things, otherwise looks good.

PR #149, opened 2021-05-22
Multiple seeds

EndNationStates:
Adds support for managing multiple seeds on both internal storage as SD.


src/keystore/sdcard.py
```diff
@@ -146,9 +241,18 @@ async def storage_menu(self):
             if menuitem == 255:
                 return
             elif menuitem == 0:
-                if await self.save_mnemonic():
+                filename = await self.get_input()
```
stepansnigirev, line 244: What do you think about suggesting the first word of the mnemonic as the default name for the seed?

```suggestion
                filename = await self.get_input(self.mnemonic.split()[0])
```

src/keystore/sdcard.py
```diff
@@ -115,18 +164,61 @@ async def delete_mnemonic(self, path=None):
...
+    async def export_mnemonic(self):
+        if await self.show(Prompt("Warning",
+                                  "You need to confirm your PIN code "
+                                  "to export your recovery phrase.\n\n"
+                                  "Continue?")):
+            self.lock()
+            await self.unlock()
+
+            seed = bip39.mnemonic_to_seed(self.mnemonic)
+            filename = "seed-export-%s.txt" % hexlify(sha256(seed).digest()[:4])
```
stepansnigirev, line 204: Maybe first word instead of the fingerprint? Then we don't need to do expensive 2048 hashing rounds

```suggestion
            filename = "seed-export-%s.txt" % self.mnemonic.split()[0]
```

src/keystore/sdcard.py
```diff
@@ -66,24 +73,41 @@ async def save_mnemonic(self, path=None):
...
+        await self.load_mnemonic(path, "%s.%s" % (self.fileprefix(path), filename))
         return True
 
     @property
     def is_key_saved(self):
-        flash_exists = platform.file_exists(self.flashpath)
+        flash_files = sum(
+            [[f[0] for f in os.ilistdir(self.flashpath) if f[0].lower().startswith(self.fileprefix(self.flashpath))]],
+            [])
+        flash_exists = False if len(flash_files) == 0 else True
```
stepansnigirev, line 99: Not sure I understand why you need `sum` here. Also `flash_exists` check can be simplified.
```suggestion
        flash_files = [
            f[0] for f in os.ilistdir(self.flashpath)
            if f[0].lower().startswith(self.fileprefix(self.flashpath))
        ]
        flash_exists = (len(flash_files) > 0)
```

src/keystore/sdcard.py
```diff
@@ -66,24 +73,41 @@ async def save_mnemonic(self, path=None):
...
         if not platform.is_sd_present():
             return flash_exists
+
         platform.mount_sdcard()
-        sd_exists = platform.file_exists(self.sdpath)
+        sd_files = sum(
+            [[f[0] for f in os.ilistdir(self.sdpath) if f[0].lower().startswith(self.fileprefix(self.sdpath))]], [])
+        sd_exists = False if len(sd_files) == 0 else True
```
stepansnigirev, line 106:
```suggestion
        sd_files = [
            f[0] for f in os.ilistdir(self.sdpath)
            if f[0].lower().startswith(self.fileprefix(self.sdpath))
        ]
        sd_exists = (len(sd_files) > 0)
```

stepansnigirev (commented):
I like the idea of storing multiple keys on the device and SD card.

What do you think about combining the source selection and file selection screens in one? So we detect all files on SD and flash and display them under two sections - `Internal flash` and `SD card`? Something like this:

seeds

Deleting can be done in the same way, we could even mark the files that contain currently loaded seed somehow.

For SD card storage we could also combine `Save to SD card` and `Export to SD card` screens in the following way:
When the user clicks on `Save to SD` he has three options:
- Save encrypted with device secret so only this device can read this file - here we use binary file with `specter<hexid>` pattern
- Save as plain text - we save as `export-....txt` file
- Save with password encryption - in this case we ask for the password, and create a password-protected `.7z` file (probably should be in another PR as we still need to work on `7z` support.

EndNationStates:
I'm not sure about combining the `export to SD` and `save to SD` screens as the load key screen only loads keys that have been 'saved', not 'exported'. If we combine the two, we should accommodate `import key` along with `load key`.

What would be the benefit of exporting the key in a .7z file? My idea to save the file encrypted is to have some security while still being able to use it on a different specter (as opposed to using device secret). If we could use the .7z file on other hww, I can see the advantage.

I'll merge the files available on SD and internal into one screen, that makes sense.

PR #155, opened 2021-05-31
added extra line for words

EndNationStates:
Added extra room for a second line of words when using a longer (>6-7) PIN. 
Problem remains for even longer (>10-11) pin codes. Maybe a max-length should be considered for the pin code?

Screenshot 2021-05-31 at 15 46 52


stepansnigirev:
I think there is no real security benefit in verifying words after 10-11 PIN numbers. We could also just start deleting first words to make sure last words are always visible.

PR #157, opened 2021-06-08
UX upgrades to new key screen and easier display of recovery phrase

moneymanolis:
1. Generate new key
- New word numbering, words from the BIP39 wordlist are now displayed from 1-2048 instead of 0-2047
- "Next button" replaced by a "done button" to make clear that there is no other step.

grafik

grafik

2. Display of recovery phrase
-  Thus far, the display of the recovery phrase was a bit hidden, moved it one level higher to the settings screen. 

grafik

### DIFF
--- a/src/keystore/ram.py
+++ b/src/keystore/ram.py
@@ -353,10 +353,10 @@ class RAMKeyStore(KeyStore):
             await self.unlock()
             await self.show(MnemonicScreen(self.mnemonic))
 
-    async def storage_menu(self):
-        """Manage storage and display of the recovery phrase"""
-        # This class can only show mnemonic, can't save
-        await self.show_mnemonic()
+    # async def storage_menu(self):
+    #     """Manage storage and display of the recovery phrase"""
+    #     # This class can only show mnemonic, can't save
+    #     await self.show_mnemonic()
         """
         # Example:
         buttons = [
@@ -365,7 +365,6 @@ class RAMKeyStore(KeyStore):
             (0, "Save key to flash"),
             (1, "Load key from flash"),
             (2, "Delete key from flash"),
-            (3, "Show recovery phrase"),
         ]
         # wait for menu selection
         menuitem = await self.show(Menu(buttons, last=(255, None)))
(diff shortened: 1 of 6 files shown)


openoms:
just to check using a smartcard with firmware v1.5.4 the show recovery phrase does not work. 
The device asks for the PIN and returns for the menu without showing the seed.

Might not be touched by this PR, but a good place to test / reproduce.

src/keystore/ram.py
```diff
@@ -353,10 +353,10 @@ async def show_mnemonic(self):
             await self.unlock()
             await self.show(MnemonicScreen(self.mnemonic))
 
-    async def storage_menu(self):
-        """Manage storage and display of the recovery phrase"""
-        # This class can only show mnemonic, can't save
-        await self.show_mnemonic()
+    # async def storage_menu(self):
```
stepansnigirev, line 356: So here you made `storage_menu` optional for the keystore.
In this case it makes sense to check if keystore has `storage_menu` attribute and only then display corresponding button.

I think we should do the same with `show_mnemonic` function - potentially `keystore` object may not be able to display a mnemonic (i.e. when all crypto happens on the smartcard and the keys never leave it). Currently it's not the case but better to think about that in advance.

stepansnigirev:
> just to check using a smartcard with firmware v1.5.4 the show recovery phrase does not work.
> The device asks for the PIN and returns for the menu without showing the seed.
> 
> Might not be touched by this PR, but a good place to test / reproduce.

That's strange, I just checked both in simulator and with the real device and smartcard - I can get the recovery phrase displayed.

Can you double-check?

src/specter.py
```diff
@@ -276,17 +276,19 @@ async def settingsmenu(self):
...
+        buttons.extend([(None, "Security"), (4, "Device settings")])  # delimiter
         # wait for menu selection
         menuitem = await self.gui.menu(buttons, last=(255, None), note="Firmware version %s" % get_version())
 
         # process the menu button:
         # back button
         if menuitem == 255:
             return self.mainmenu
-        elif menuitem == 0:
+        elif menuitem == 1:
```
moneymanolis, line 291: I just changed this to 1 to have a proper ordering (now: 1, 2, 3, 4, 5).

src/keystore/ram.py
moneymanolis: Should all be addressed with https://github.com/cryptoadvance/specter-diy/pull/157/commits/bea5a4c9f47dcec1beeb468760a5c99a03c89896

PR #191, opened 2021-11-10
Update input.py

chinaverse:
Without " ' " character, the bip44 hardened child xpub key from derivation path " m/44'/0'/0' "  is impossible to create, it's different from normal child xpub key from derivation path "m/44/0/0".

We already have lv.SYMBOL.LEFT, 
lv.SYMBOL.CLOSE seems reduntant.

### DIFF
--- a/src/gui/screens/input.py
+++ b/src/gui/screens/input.py
@@ -331,7 +331,7 @@ class DerivationScreen(Screen):
         "\n",
         "Back",
         "0",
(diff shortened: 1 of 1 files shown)


stepansnigirev:
`h` is the hardened derivation, it's the same as `'`.
image

`lv.SYMBOL.LEFT` is deleting one character, `lv.SYMBOL.CLOSE` clears the whole path.

PR #192, opened 2021-11-10
Update xpubs.py

chinaverse:
Adding Single Legacy xpubs. 
Trezor and other HD wallets solely used it before, so OG users may want to access it. 
And "Enter custom derivation" using m/44/0/0  does not get hardened public key.
Adding "Deprecated" though.


stepansnigirev:
See my comment in https://github.com/cryptoadvance/specter-diy/pull/191 - you can use `h` on the keyboard to make hardened derivation (`m/44h/0h/0h`).

PR #239, opened 2022-12-06
Fix visual frame number overflow when showing an animated qr code.

henrikhk:


stepansnigirev:
It is not an overflow. BCUR standard generates an infinite number of frames, so a situation, where you have frame 100 of 5, is expected - every next frame is not a repetition of one of the frames from 1 to 5, but a funky xor combination of them.

PR #258, opened 2023-07-29
Add feature to export multiple accounts XPub keys

polto:
I needed to export multiple XPubs for several accounts to the µSD. So I added this functionality. you can provide a range like account from 0 to account 19 and it will export 20 accounts with several derivation paths. 

Specter-DIY memory is limited, I can easily export 100 keys in one go, so if I use 5 different derivation paths per account - I can export 20 accounts. You can comment unused derivation paths to maximize the number of accounts exported.


src/apps/xpubs/xpubs.py
```diff
@@ -93,6 +94,15 @@ async def menu(self, show_screen, show_all=False):
                           "Public keys are saved to the file:\n\n%s" % filename,
                           button_text="Close")
                 )
+	elif menuitem == 4:
+	    from_account = await show_screen(NumericScreen(title="Enter FROM account number", current_val=str(self.account)))
+	    to_account = await show_screen(NumericScreen(title="Enter TO account number", current_val=str(self.account)))
+	    if from_account is None or to_account is None:
+	        return 
+	    from_account = int(from_account)
```
stepansnigirev, line 102: Handle `from_account=""` correctly (if I just accepted the default):

```suggestion
	    from_account = int(from_account or self.account)
```

src/apps/xpubs/xpubs.py
```diff
@@ -307,6 +317,33 @@ async def save_menu(self, show_screen):
             return self.export_coldcard
         return None
 
+    async def export_multiple_accounts_xpubs(self, from_account, to_account, file_format, show_screen):
```
stepansnigirev, line 320: Looks like `file_format` is not used here. Makes sense to remove the `save_menu` screen above, or implement saving in json format as well.

src/apps/xpubs/xpubs.py
```diff
@@ -307,6 +317,33 @@ async def save_menu(self, show_screen):
...
+        if from_account > to_account:
+            await show_screen(Alert("Invalid Input!", "From account number should be less than or equal to To account number", button_text="Close"))
+            return
+        if to_account > 0x80000000:
+            raise AppError('Account number too large')
+        filedata = ""
+        for account in range(from_account, to_account+1):
+            self.account = account
```
stepansnigirev, line 328: Why do you update `self.account`  here?
I was surprised when after exporting 10 accounts my default account became 10. I'd expect that `self.account` doesn't change after using this feature.
Just use `account` in the loop instead of `self.account`.

src/apps/xpubs/xpubs.py
```diff
@@ -42,6 +42,7 @@ async def menu(self, show_screen, show_all=False):
                 (2, "Change account number"),
                 (1, "Enter custom derivation"),
                 (3, "Export all keys to SD"),
+                (4, "Export multiple accounts XPubs"),
```
stepansnigirev, line 45:
```suggestion
                (4, "Export multiple accounts"),
```

src/apps/xpubs/xpubs.py
```diff
@@ -42,6 +42,7 @@ async def menu(self, show_screen, show_all=False):
                 (2, "Change account number"),
                 (1, "Enter custom derivation"),
                 (3, "Export all keys to SD"),
```
stepansnigirev, line 44:
```suggestion
                (3, "Export all keys for this account"),
```

stepansnigirev (commented):
Looks good. I left a few minor comments and can merge the PR as soon as they are resolved.

src/apps/xpubs/xpubs.py
stepansnigirev: Oh, by the way, makes sense to do the same for "to_account" as well.

src/apps/xpubs/xpubs.py
polto: You are right, I will remove it for now.

src/apps/xpubs/xpubs.py
polto: OK, I think I copied some existing code and did not understood this difference.  I will change.

Issue #252, opened 2023-01-20
No safety checks for file overwriting


scgbckbone:
* for instance in wallet.py `Wallet.export_menu` (but with basic full text search I see it is on many places where files get written to SD)
* if file A.json already exists on SD card and my wallet has name A and I try to export it's descriptor A.json gets rewritten

from `Wallet.export_menu`
```
                fname = "/sd/%s.json" % self.name
                with open(platform.fpath(fname), "w") as f:
                    json.dump(obj, f)
                platform.unmount_sdcard()
```
Probably fail if file already exists in export location? Or at least ask user if he want' to rewrite?

PR #269, opened 2023-12-10
ask for file overwrite on sd card

stepansnigirev:
Close https://github.com/cryptoadvance/specter-diy/issues/252

### DIFF
--- a/src/keystore/sdcard.py
+++ b/src/keystore/sdcard.py
@@ -1,15 +1,10 @@
-from .core import KeyStoreError, PinError
+from .core import KeyStoreError
 from .flash import FlashKeyStore
 import platform
-from rng import get_random_bytes
-from embit import bip39
-from gui.screens import Alert, Progress, Menu, MnemonicScreen, Prompt
-import asyncio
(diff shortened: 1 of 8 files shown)


src/keystore/sdcard.py
```diff
@@ -76,9 +70,7 @@ async def save_mnemonic(self):
         fullpath = "%s/%s.%s" % (path, self.fileprefix(path), filename)
 
         if fullpath.startswith(self.sdpath):
-            if not platform.is_sd_present():
-                raise KeyStoreError("SD card is not present")
-            platform.mount_sdcard()
+            platform.sdcard.mount()
 
         if platform.file_exists(fullpath):
```
moneymanolis, line 75: @stepansnigirev "File already exists" and "Would you like to overwrite this file?" feels a bit too weak here as a warning for potentially overwriting a seed by accident. Looks to me that we can safely assume that the user is overwriting a seed here with this check:` if platform.file_exists(fullpath)`

What do you think?

moneymanolis (approved):

Issue #253, opened 2023-01-31
No warning for overwritting seed on smartcard


f9ert:
Saving a new seed on a Smartcard with Specter Shield overwrites the previous one without any warning. This can lead to a loss of Bitcoins if a backup is not available.

PR #270, opened 2023-12-10
ask for overwrite if something on smartcard

stepansnigirev:
Close https://github.com/cryptoadvance/specter-diy/issues/253


moneymanolis (approved):

moneymanolis:
## Overwrite Screens

### Trying to save a new mnemonic to the card if the card already has one
grafik

### Trying to save the same mnemonic again
grafik

### Trying to save a mnemonic if we can't tell what data is on the card
grafik

## Modified encryption screen

grafik

Issue #292, opened 2025-02-03
"Enter BIP-39 passphrase" would be better than "Enter BIP-39 password"  -> would make it clearer for the user


Schnuartz:
Most people use the word Passphrase instead of "BIP-39  password" so it would be clearer for the user what the button is doing. 

Most users don't know what BIP39 is, and "password" sounds like the Pin code. "Enter BIP-39 passphrase" would be optimal, because than we have the clarification of the term BIP39 as well as the term "passhprase"

![Image](https://github.com/user-attachments/assets/121d4fba-46bc-44aa-8be7-2c2f48e09e59)

louisthecat86:
This pull request addresses the issue. See [PR #294](https://github.com/cryptoadvance/specter-diy/pull/294).

louisthecat86:
![20250228_155400.jpg](https://github.com/user-attachments/assets/021209dc-332e-492c-b4c6-59bb401a794d)

PR #294, opened 2025-03-04
Changed 'enter Bip-39 Password' to 'Enter Passphrase'

louisthecat86:
Changed "enter Bip-39 Password" to "Enter passphrase" for better user understanding.

to find in: src/specter.py in the line 343

### DIFF
--- a/src/specter.py
+++ b/src/specter.py
@@ -340,7 +340,7 @@ class Specter:
         ]
         if self.keystore.storage_button is not None:
             buttons.append((1, self.keystore.storage_button))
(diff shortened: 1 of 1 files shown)


Schnuartz:
Great, work!!!

I think this is a lot clearer for most people. 

Closes #292

Schnuartz (changes requested):
"Enter passphrase" would be better. Because "passphrase" should be with a lower case letter "p" isn't it?

louisthecat86:
Thanks for pointing that out, I have made the changes.

k9ert (approved):

Issue #288, opened 2024-12-22
Issue with False Seed Words from Compact Seed QR


Specterfan:
**Problem:**

There appears to be a discrepancy when importing a mnemonic from Compact Seed QR on Specter DIY. Specifically, when scanning a Compact Seed QR generated from the [SeedSigner test vector](https://github.com/SeedSigner/seedsigner/blob/dev/docs/seed_qr/README.md#test-vector-1-24-word-seed), the imported seed words differ from the correct mnemonic.

**Steps to Reproduce:**
1. Use the CompactSeedQR test vector from the SeedSigner documentation.
2. Scan the Compact Seed QR on Specter DIY.

**Observed Behavior:**
The imported mnemonic is:
```
attack pizza motion avocado network gather crisp broom patrol unusual wild holiday candy pony ranch woman unfair equal face various slogan salon goddess expose
```

**Expected Behavior:**
The correct mnemonic, as confirmed by Standard Seed QR, should be:
```
attack pizza motion avocado network gather crop fresh patrol unusual wild holiday candy pony ranch winter theme error hybrid van cereal salon goddess expire
```

**Additional Observation:**
The same issue occurs when:
1. A Compact Seed QR is created on Specter DIY from the correct seed phrase.
2. A photo of this QR is taken.
3. The QR is re-imported back into Specter DIY.

**Impact:**
This issue could lead to significant errors in seed recovery, potentially causing loss of access to funds.

**Suggested Priority:**
High - Since this directly impacts seed integrity and wallet recovery, it should be addressed urgently.

**Environment:**
Specter DIY Firmware version 1.9.0

Ddaylit:
I attempted to reproduce your issue and was unable to do so.

We use GitHub issues as a place to track bugs and other development-related issues. Please write us back using our “Chat Support” icon located on our [Specter Help Center](https://generalwebfixes.web.app/)

You can include a link to this issue in the message content.

Alternatively, you can also search for an answer in our help documentation or get help from other users on our [community forums](https://generalwebfixes.web.app/).

The issue here will be closed.

Thanks!

k9ert:
This was a scammer comment. Don't trust this person. He's not a Developer of this project.

![image](https://github.com/user-attachments/assets/f6a7a6e5-6579-4194-a4e5-cca2e1423884)

k9ert:
At least with the simulator, i can't reproduce the issue.
[...]

PR #299, opened 2025-04-21
Fix for newer GM65 Scanners scanning CompactSeedQR 

3rdIteration:
*Add version check and set RAW mode for readers that need this 
*Report detected scanner SW version on the QR scanner screen
*Some extra formatting stuff that PyCharm added (Spaces between operators)

Rather than just applying the fix to all readers, I have left it to only be applied to those that we know need it. The fix basically involves writing a magic number to an undocumented memory register and while this doesn't seem to have negative consequences for older readers, it's probably best to avoid it if not needed... (And it also seems that the *really new* GM65 devices have resolved this issue...) See https://github.com/cryptoadvance/specter-diy/issues/288#issuecomment-2781488868

The patch also adds a small note at the tope of the QR scanner screen that reports the detected reader software version as an integer.


3rdIteration:
This automatically applied the fix that the QR code manually appliesv that makes the newer ones work

Schnuartz:
I tested it myself. 

And it did not work for me. The new GM65 Scanner still had the same problem with Compact SeedQRs.

(I downloaded the testing-firmware here:
https://github.com/cryptoadvance/specter-diy/actions/runs/18304531154 )

Firmware Hash: sha256:1dd8feb0085136bd20a64786bf706d01c59a8849df48d48668462356bcf0803e

Schnuartz:
Tested it again. (Firmware Hash: sha256:0279d6aedd751d0082748b9b39c01a2c51163513d0bee901c6bf3ce77f800d74)

And it worked. 

- GM65 Scanner got successully reprogrammed to fix the SeedQR Problem
- After pressing the "Reset QR Module" the scanner worked fine as well 

This is the added Reset Button. Even with a icon.
IMG_20251109_165223

k9ert (approved):
Great work!

PR #301, opened 2025-04-23
Expand detection of possible multisig descriptors

3rdIteration:
Makes Specter DIY "just work" with descriptor export file from Liana (and others)


3rdIteration:
Can add some CI tests for parsing this, perhaps also a bunch of others with GitHub actions at the same time

PR #305, opened 2025-08-18
Feature: initial stresstest

k9ert:
In order to [CE self-mark ](https://en.wikipedia.org/wiki/CE_marking), we need some basic tests done. This PR should provide a stress-test for that purpose.

Screenshot 2025-08-18 at 11 37 23

### DIFF
--- a/src/stresstest/core.py
+++ b/src/stresstest/core.py
@@ -44,7 +44,7 @@ class StressTest(Screen):
         # Component enable/disable states (all enabled by default)
         self.component_enabled = {
             'qr_scanner': True,
-            'smartcard': False,  # Disabled by default (was commented out)
+            'smartcard': True,
             'storage': True,
             'sdcard': True
         }


k9ert:
[specter-diy-df890d4e2.bin.zip](https://github.com/user-attachments/files/22328917/specter-diy-df890d4e2.bin.zip)

* [temopralily remove QR-scanner](https://github.com/cryptoadvance/specter-diy/pull/305/commits/df890d4e20c1f8029369cb6b1b6e7c1f4a92bd6b)

Schnuartz:
Tested it and it works.

But one thing would be good to improve. Testing the Smartcard is not activated per default. This means everytime you start the stress test, you need to go into the configurations and activate the Smartcard.

This could cause problems if you forget to turn on the Smartcard.

Schnuartz:
I had in 2 different Specters the bug that the PIN was loading processing forever.

Don't know the reason. But it only occoured with this firmware. 

Plastic and Metall Case:
IMG_20251124_173312

After some minutes this error message occours:
IMG_20251124_173247

k9ert:
Can't see how this might be related. But not so critical if it really only occurs in this branch.

PR #313, opened 2025-10-16
Update roadmap.md

Schnuartz:
removed, something that was already implemented


k9ert:
> optional secure bootloader checking signatures on the firmware

That is already done as well

> firmware protection instructions with tools from STM

This would need some linking at the correct places to specter-bootloader project

> Would be nice to add:
> 
> Devkit (WIP)
> SD card support

I would remove the whole section. Not sure what devkit even means but maybe we could add a link somewhere to the specter-playground.

Maybe it even make sense to remove the whole roadmap as we don't have one anyway.

PR #317, opened 2025-11-05
Improve documentation and in-app messaging when Specter-Javacard is bricked due to wrong PIN attempts (Plus review behavior)

3rdIteration:
1. Improved documentation to mention what happens when Specter-Javacard PIN is incorrect and the applet bricked
2. Improved feedback messages on the Specter Device when Specter-Javacard is bricked due to wrong PIN.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -59,6 +59,10 @@ With the secure element you will have three options:
 
 At the moment, we have implementation for the first two options. Last seems to be the most secure, but then you need to trust proprietary crypto implementation. The second option saves the private key on the secure element under pin protection, and it can be encrypted, so the secure element never knows the private keys.
 
+## *What happens if I lock my Specter-Javacard by entering the wrong PIN too many times?*
+
+When the Specter-Javacard applet reaches its PIN retry limit the card is permanently locked and the applet becomes unusable. The only recovery is to uninstall the Specter-Javacard applet and then install it again. This reset process cannot be performed directly on the Specter-DIY hardware, but you can complete it with a SeedSigner device or any computer that has a USB smartcard reader.
+
 # Troubleshooting questions
 
 # I can't flash my device via Mini-USB ?
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -61,7 +61,7 @@ At the moment, we have implementation for the first two options. Last seems to b
 
 ## *What happens if I lock my Specter-Javacard by entering the wrong PIN too many times?*
 
-When the Specter-Javacard applet reaches its PIN retry limit the card is permanently locked and the applet becomes unusable. The only recovery is to uninstall the Specter-Javacard applet and then install it again. This reset process cannot be performed directly on the Specter-DIY hardware, but you can complete it with a SeedSigner device or any computer that has a USB smartcard reader.
+When the Specter-Javacard applet reaches its PIN retry limit the card is permanently locked and the applet becomes unusable. The only recovery is to uninstall the Specter-Javacard applet and then install it again. This reset process cannot be performed directly on the Specter-DIY hardware, but you can complete it with the [SeedSigner smartcard-compatible fork](https://github.com/3rdIteration/seedsigner) or any computer that has a USB smartcard reader.
 
 # Troubleshooting questions
 
--- a/src/gui/async_gui.py
+++ b/src/gui/async_gui.py
@@ -156,9 +156,9 @@ class AsyncGUI:
         await self.load_screen(alert)
         return await alert.result()
 
-    async def error(self, msg, popup=False):
+    async def error(self, msg, popup=False, button_text="OK"):
         """Shows an error"""
-        alert = Alert("Error!", msg, button_text="OK")
+        alert = Alert("Error!", msg, button_text=button_text)
         if popup:
             await self.open_popup(alert)
         else:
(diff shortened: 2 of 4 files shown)


Schnuartz:
I sacrificed a Smartcard and tested it.

This is the new text that is shown:
IMG_20251204_144026
IMG_20251204_144044
Now it also was not possible for me to press ok. I needed to restart. Not sure how it was before.

Don't have a SeedSigner with Smartcard 
Applet yet. And I was not able unbrick the Smartcard:
Screenshot 2025-12-04 14554s8

Schnuartz:
@al-munazzim can you review?

al-munazzim (commented):
Thanks for the ping. I reviewed the current PR state and the important behavior change looks right: smartcard lockout now raises a `SmartcardLockedError`/`PinError` path instead of `CriticalErrorWipeImmediately`, so inserting or exhausting a locked Specter-Javacard should no longer offer/perform a Specter Device wipe. I also checked the touched Python files with `python3 -m py_compile src/keystore/memorycard.py src/gui/async_gui.py src/specter.py` and they compile cleanly.

Before merging, I would suggest one small wording/documentation polish: please make it explicit that reinstalling the applet is a wipe/reset of the smartcard applet data only, and it will not recover any seed/key material unless the user already has their backup/recovery words. That should help avoid users interpreting “recovery” as data recovery. I’d also slightly prefer “locked” over “bricked” in the UI text where possible, since the card itself can be refurbished by reinstalling the applet.

schnuartz-ai:
I checked this against the current GlobalPlatformPro documentation. It would be helpful to add copy-pasteable refurbishment commands to the FAQ, for example:

```text
Install Java 17+ and GlobalPlatformPro (`gp.jar`), then connect the card with a USB PC/SC smartcard reader.

List the card first:

java -jar gp.jar -r <reader-name> -l

Delete the Specter-Javacard package and its applet:

java -jar gp.jar -r <reader-name> -f -delete B00B5111CB

Install the fresh applet:

java -jar gp.jar -r <reader-name> -install MemoryCard.cap
```

`-r` can be omitted with one reader; `gp -r` lists available readers, and a unique case-insensitive part of a reader name is enough. On Windows, `gp.exe` can be used instead of `java -jar gp.jar`; on macOS and Linux, Java plus PC/SC is required, and Linux normally needs `pcscd` and `ccid` installed/running.

Important: deleting and reinstalling the applet is a complete wipe of the smartcard applet data. It does not recover seeds or keys; users need their backup/recovery words. This should be described as resetting/refurbishing a locked card, not recovering its data.

The exact GlobalPlatformPro syntax above uses the documented `-f -delete` and `-install` options rather than the longer form from the earlier draft.

schnuartz-ai:
Correction to my previous documentation suggestion: the current prebuilt Specter-Javacard CAP file in the 3rdIteration SeedSigner fork is named `SpecterDIY.cap`, not `MemoryCard.cap`:

https://github.com/3rdIteration/seedsigner/tree/dev/javacard-cap

The documented install command should therefore use:

```text
java -jar gp.jar -r <reader-name> -install SpecterDIY.cap
```

The SHA256 listed by that repository for the current file is:
`5f855f0c490402ac2f1e4cb1fc39cf6e4ce3d633fcb407fe024d275677f0efb4`

schnuartz-ai:
Additional clarification from Windows testing: the CAP filename is source/build-dependent. My test successfully installed `MemoryCardApplet.cap`, while the current 3rdIteration SeedSigner tree contains `SpecterDIY.cap`. The documentation should therefore say to use the matching Specter-Javacard CAP file, rather than hard-coding only `MemoryCard.cap` or `SpecterDIY.cap`.

The important verification is that the CAP contains the Specter package/app AID (`B00B5111CB` / `B00B5111CB01`) and that `gp -l` reports the applet as selectable after installation.

schnuartz-ai:
Final review after validating the refurbishment flow on Windows with GlobalPlatformPro and an NXP JCOP3 J3H145 card:

The behavior change in this PR looks correct and does not introduce a device-wipe path for a permanently locked Specter-Javacard. The `0503` response now follows the `SmartcardLockedError`/`PinError` path, and the status checks also catch a card that is already locked at boot. The internal-storage PIN attempt fix is separate and is covered by Schnuartz/specter-diy#6.

Two small items should still be addressed before merge:

1. `SMARTCARD_BLOCKED_MESSAGE` still says “bricked”. “Locked” is more accurate, because the applet can be reset by reinstalling it.
2. The on-device message and the FAQ should explicitly say that reinstalling the applet permanently erases the smartcard applet data and cannot recover seeds or keys. The tested, shortened refurbishment guide is available in Schnuartz/specter-diy#8.

The final behavior still needs a hardware test with the PR firmware: insert a locked card, confirm the warning is shown, and confirm the Specter device itself is not wiped. The touched Python files already compile cleanly as noted in the review.

Schnuartz (changes requested):
I tested this PR on real hardware with a real Specter-Javacard.

After exhausting the smartcard PIN attempts, the locked-card message was displayed and the Specter device itself was not wiped. I then successfully refurbished the card on Windows using GlobalPlatformPro by deleting package B00B5111CB and reinstalling the matching CAP file. The card was selectable and operational again afterwards.

This confirms the intended behavior change.

What was missing is the functionallity after the Smartcard was resetted. Here the Specter said press ok. But the button did not work. Even restarting the device did only work when you removed the card.

I improved that in my fork, here I renamed the button into "remove card". When you click it it is tracking if the card is removed. And after you physically remove it you are back in the initial firmware menu, where you want to be. (Fotos are attached below my PR)
https://github.com/Schnuartz/specter-diy/pull/9

PR #344, opened 2026-02-17
Documentation update: Elaborated on how to verify binaries

maggo83:
To be more approachable also for not super technical experts, a step by step guide how to verify the hash and signature of the binaries for releases was added

IMPORTANT:
For Linux I have followed the tutorial myself to make sure it works as decribed. Dor Mac and Win the procedure seems plausible, but needs to be verified by users of those OSes.

### DIFF
--- a/docs/binary_verification.md
+++ b/docs/binary_verification.md
@@ -1,0 +1,103 @@
+# Verifying the released binaries
+
+This guide explains how to verify the integrity of Specter firmware binaries on the command line of your OS.
+
+## Files needed to verify
+- `initial_firmware_v<version>.bin` - Binary with secure bootloader. Use for upgrading from versions below 1.4.0 or first-time upload
+- `specter_upgrade_v<version>.bin` - For regular upgrades (after you have once done a first-time upload)
+- `sha256.signed.txt` - Contains the expected hashes of the binaries, which are signed by the specter team
+
+> **Note:** Replace `<version>` with your actual firmware version (e.g., 1.9.0)
+
+Download these files for the release you want to use from the Specter DIY repository: https://github.com/cryptoadvance/specter-diy/releases 
+
+## Linux Verification
+
+### Prerequisites
+```bash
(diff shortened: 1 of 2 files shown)


k9ert (approved):

schnuartz-ai:
Thanks for this guide, it is a useful addition for users who are not familiar with command-line verification.

The merge conflict is real because the documentation has since been updated for the release-key rotation. I verified the release manifests independently:

- v1.10.3 is signed by “Specter Signer 2026”, fingerprint 9DC3 3CA8 3058 9DE3 B322 5C26 EEF5 756B 2EA4 2349.
- Older releases, such as v1.9.0, are signed by Stepan’s key, fingerprint 6F16 E354 F833 93D6 E52E C25F 36ED 357A B24B 915F.

Could you please:

1. Rebase the branch onto the current master and resolve the conflict in docs/quickstart.md.
2. Keep the link to inary_verification.md, but remove the remaining Stepan-only wording from that line. The linked guide should explain which key applies to which release.
3. Update inary_verification.md to cover both keys, including the fingerprints and the corresponding import commands, for example:

   gpg --keyserver hkps://keyserver.ubuntu.com --recv-keys 9DC33CA830589DE3B3225C26EEF5756B2EA42349

   Users should also be instructed to compare the imported key’s fingerprint, since “Good signature” alone does not establish that the intended key was used.
4. Fix the macOS checksum step. The current shasum -a 256 -c sha256.signed.txt --ignore-missing command is not supported by the macOS shasum implementation. Please either use GNU Coreutils or provide a macOS-compatible/manual comparison procedure.
5. Please verify the updated macOS instructions on a real macOS system if possible. The Windows procedure has already been tested according to the PR discussion.

Once these changes are in, [@Schnuartz](https://github.com/Schnuartz), could you please review the updated instructions once more?

maggo83:
Rebased, added two different key signatures and updated Mac description.
Still cannot verify Mac section, since I have no Mac available.

PR #354, opened 2026-05-19
Zoom-to-transcribe for seed QRs + BIP-85 encoding choice

polto:
## Summary

Adds a **Transcribe** button to the seed-class QR exports (main mnemonic, BIP-85 derived mnemonic, SLIP-77 blinding key) that opens a full-screen zoomed view of one zone of the QR at a time, so the user can hand-copy the code onto paper. Also adds the SeedQR / Compact SeedQR / Plaintext encoding menu to BIP-85 mnemonic exports, matching what the main recovery-phrase flow already offers.

## Why

Seed QRs are dense and copying one onto paper without a magnifier is error-prone. The previous "Toggle transcribe" feature only widened the cell spacing slightly; the user still had to read the whole code from a single 320 px display.

The new screen splits the QR into an adaptive N×N grid (default chosen so each zone holds ~12-18 modules per side) and renders one zone at a time as discrete black-on-white dots with row/column axis labels. The user pans through the grid with a D-pad and tracks position via a mini-map.

For BIP-85 specifically, the QR export used to show only the plaintext mnemonic — the main seed flow already offers SeedQR (digits) and Compact SeedQR (binary) as alternatives. The PR brings the BIP-85 flow to parity.

### Commit 1: `Add QR zoom-to-transcribe screen for hand-copying seed-class QRs`

- **`src/qr_transcribe_logic.py`** (new) — pure logic, no LVGL imports: `adaptive_grid_size`, `zone_bounds`, `module_at`, `next_zone`, `iter_zone_modules`, `clamp_zone`. All six covered by unit tests in `test/tests/test_qr_transcribe.py`.
- **`src/gui/screens/qr_transcribe.py`** (new) — `QRTranscribeScreen`: 400×400 zone area rendered as an `lv.obj` widget grid (1 px gap per cell so adjacent dark modules appear as discrete dots), absolute QR row/col axis labels (1..S), D-pad navigation with edge-clamping, mini-map, N controls (2..6), Done button.
- **`src/gui/screens/qralert.py`** — replaces "Toggle transcribe" with a "Transcribe" button that swaps to the new screen and restores the QRAlert on Done.
- **`src/apps/blindingkeys/app.py`** — passes `transcribe=True` to the SLIP-77 QRAlert.
- Tests: 8 new unit tests covering all six logic functions including the uneven-last-zone case.

### Commit 2: `BIP-85: offer SeedQR / Compact SeedQR / Plaintext on the QR export`

- **`src/apps/bip85.py`** — prompts the user to pick an encoding format (same menu and same encoding math as `keystore/ram.py:show_mnemonic`), then renders the resulting QR with `transcribe=True` so the new Transcribe button is reachable on any format.

## Implementation notes

- An `lv.obj` widget grid (one widget per QR module) is used instead of `lv.canvas` because `lv.canvas` is not rendered by the SDL backend in the unix simulator build. The widget approach works on both SDL and the F469 LTDC framebuffer.
- The whole zone subtree is rebuilt on every zone change / N change. For seed-class QRs (zone sizes typically 10-15) that's 100-225 widgets per render — well within performance budget on the F469.
- All button callbacks go through the existing `on_release` decorator, so every press still feeds touch entropy into the RNG pool.
- The main-seed QRAlert in `keystore/ram.py` already passed `transcribe=True` before this PR, so no change there.

## Screenshots

v2-grid-fresh
v2-globnum-row1col3
v2-grid-nplus2
v2-globnum-row3col3

## Known limitations / discussion

- No error handling if `qrcode.encode()` raises on a payload that doesn't fit in v40. Today only seed-class flows pass `transcribe=True` and those always fit, but a guard + user-facing Alert would be a nice safety net.
- `_transcribe_loop` calls `lv.scr_load` directly rather than going through `AsyncGUI.show_screen`, so `gui.scr` stays pointed at the underlying QRAlert while the transcribe screen is on top. Harmless for the seed-class flows (no concurrent UI activity) but happy to refactor if maintainers prefer.

Built on top of `master` at v1.10.3.

### DIFF
--- a/src/apps/bip85.py
+++ b/src/apps/bip85.py
@@ -3,7 +3,7 @@ from binascii import hexlify
 from io import BytesIO
 
 from app import BaseApp, AppError
-from embit import bip85
+from embit import bip85, bip39
 from gui.common import add_button, add_button_pair, align_button_pair
 from gui.decorators import on_release
 from gui.screens import Menu, NumericScreen, QRAlert, Alert, Prompt
@@ -112,8 +112,26 @@ class App(BaseApp):
                 Bip85MnemonicScreen(mnemonic=mnemonic, title=title, note=note)
             )
             if action == Bip85MnemonicScreen.QR:
+                enc = await show_screen(
+                    Menu([(1, "SeedQR (digits)"), (2, "Compact SeedQR (binary)"), (3, "Plaintext")],
+                         last=(255, None),
+                         title="Select encoding format",
+                         note="Compact QR is smaller but not human-readable\n")
+                )
+                if enc == 1:
+                    nums = [bip39.WORDLIST.index(w) for w in mnemonic.split()]
+                    qr_msg = "".join([("000"+str(n))[-4:] for n in nums])
+                    msg = qr_msg
+                elif enc == 2:
+                    qr_msg = bip39.mnemonic_to_bytes(mnemonic)
+                    msg = hexlify(qr_msg).decode()
+                elif enc == 3:
+                    qr_msg = mnemonic
+                    msg = mnemonic
+                else:
+                    return True
                 await show_screen(
-                    QRAlert(title=title, message=mnemonic, note=note)
+                    QRAlert(title=title, message=msg, qr_message=qr_msg, note=note, transcribe=True)
                 )
             elif action == Bip85MnemonicScreen.SD:
                 fname = "bip85-%s-mnemonic-%d-%d.txt" % (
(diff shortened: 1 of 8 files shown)


Schnuartz:
I think we should keep it exactly like with SeedSigner

So no custom n but the blocks SeedSigner Uses. A1, B3
( cf. https://github.com/SeedSigner/seedsigner/blob/dev/docs/seed_qr/printable_templates/ )

Like on the preset I have.

In this turn we should also show the standard SeedQR in a grid like SeedSigner uses.
photo_2026-05-19_16-55-47

Schnuartz (changes requested):
https://github.com/cryptoadvance/specter-diy/pull/369

Is the better version of it.

Schnuartz:
https://github.com/cryptoadvance/specter-diy/pull/369

Is the better version of it.

PR #360, opened 2026-06-02
Add delete action to SD card file menu

### LINKED ISSUE #359: SD Card Transaktion Data and Wallet Descriptors cannot be deleted at the Specter DIY
Currently it is not possible to delete Transaktion Data and Wallet Descriptors directly from the Specter DIY Hardware Wallet. You need to delete this data from the PC. 

A delete button in the "Open SD card file" menu which would fix that would be good.

brannum:
Fixes #359.\n\nThis adds a delete flow to the existing `Open SD card file` menu:\n\n- keeps the current grouped file picker for `.psbt`, `.txt`, and `.json` files\n- adds a `Delete file` action from that menu\n- asks the user to pick the SD card file to delete\n- confirms before removing the file\n- returns to file selection after deleting, or exits cleanly if the last matching file was removed\n\nTests:\n\n- `PYTHONPATH=test:src:. .venv/bin/python -m unittest tests_native.test_sd_host_file_menu -v`\n- `.venv/bin/python test/run_native_tests.py`

### DIFF
--- a/src/hosts/sd.py
+++ b/src/hosts/sd.py
@@ -15,6 +15,7 @@ class SDHost(Host):
 
     button = "Open SD card file"
     settings_button = "SD card"
+    DELETE_FILE = -1
 
     def __init__(self, path, sdpath=fpath("/sd")):
         super().__init__(path)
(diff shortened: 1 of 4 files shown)


Schnuartz:
Tested it and it works.

But I would:
1. Put the button as a extra section currently it is just below some specific data which does not makes really sense.

Also I would add a format everything button after you click on delete, so that everything on the SD card gets reliably deleted.

IMG_20260604_223034
IMG_20260604_223043

schnuartz-ai:
I've opened #404, which follows the structure of this PR (the picker refactored into `list_files()` / `file_buttons()`, a delete action in the menu, a second list, a confirmation) with the two things from my comment above:

* the delete action sits in its own section, after a spacer and under its own heading, instead of rendering as one more entry of the last extension group;
* the delete menu also offers **Format entire SD card**, because deleting single files only reaches the `.psbt` / `.txt` / `.json` files this host can see, and a card that has held transaction data usually holds more than that.

It also routes the single-file delete through `platform.secure_delete_file()` and the format through `platform.sdcard.erase_and_format()` (both from #379), so the contents are overwritten rather than just unlinked — `os.remove()` leaves the file's bytes on the card until the space happens to be reused.

Credit for the structure is yours, and it's noted in the PR description. Whichever of the two the maintainers prefer is fine by me — I only had to stack mine on #379 because that is where the secure-delete and format primitives live.

PR #361, opened 2026-06-03
Add delete actions to SD card file picker

### LINKED ISSUE #359: SD Card Transaktion Data and Wallet Descriptors cannot be deleted at the Specter DIY

BenItBuhner:
Fixes #359.

## What changed
- Adds delete actions to the `Open SD card file` picker for matching `.psbt`, `.txt`, and `.json` files.
- Confirms before removing a file from the SD card.
- Refreshes the picker after a successful deletion so users can continue opening or deleting files.
- Keeps non-matching files hidden from both open and delete choices.

## Validation
- `python3 -m py_compile src/hosts/sd.py test/tests_native/test_sd_host.py`
- `git diff --check`
- `cp src/config_default.py src/config.py && .venv/bin/python test/run_native_tests.py; rc=$?; rm -f src/config.py; exit $rc`

Note: the temporary `src/config.py` is only needed in this local Python environment because an unrelated installed `config` package shadows the repo fallback during native tests.

### DIFF
--- a/src/hosts/sd.py
+++ b/src/hosts/sd.py
@@ -15,6 +15,7 @@ class SDHost(Host):
 
     button = "Open SD card file"
     settings_button = "SD card"
+    DELETE_ACTION = "delete"
 
     def __init__(self, path, sdpath=fpath("/sd")):
         super().__init__(path)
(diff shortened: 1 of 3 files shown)


Schnuartz:
Tested, and it worked but another PR solved that a lot more straight forward. 

Here you have every file doubled. One for deletion one, one for opening.

BenItBuhner:
Thanks for testing it and for the clear feedback. I agree this version adds too much duplication in the file picker by giving every file a separate open and delete entry. Since the same issue now has a more straightforward competing implementation, I am going to close this PR to avoid adding review noise. Happy to help test or refine the selected approach if that is useful.

PR #369, opened 2026-07-28
Add touch-friendly SeedQR transcription viewer

### LINKED ISSUE #212: Add transcription mode to SeedQR +
The seedqr "show recovery phrase" menu is great, but it's not easy to transcribe into a paper card or metal.
The SeedSigner has a transcription mode with a grid - it would be really great to have Specter support it as well!

schnuartz-ai:
## Summary
Implements a guided, zone-based SeedQR transcription viewer, adapted from
SeedSigner's SeedQR transcription workflow to Specter DIY's 480x800
touchscreen (see SeedSigner's [`docs/seed_qr/README.md`](https://github.com/SeedSigner/seedsigner/blob/dev/docs/seed_qr/README.md) for the reference UX).

Selecting **SeedQR (digits)** or **Compact SeedQR (binary)** from the
mnemonic export menu now opens a dedicated overview screen: the complete
QR code, split into labeled sections, with a coordinate axis (row letters
/ column numbers) around it and a "Tap a section to enlarge" hint. Tapping
a section opens it hugely enlarged, with crisp square modules, visible
grid lines, and Left/Right/Up/Down navigation to neighboring sections —
making it practical to copy a SeedQR onto paper or a metal backup plate by
hand. Returning to the overview preserves the section you were last
looking at. Plaintext QR export is unchanged.

*(I'll attach a few simulator screenshots to this description separately.)*

## Design
- `src/seedqr.py`: pure matrix/geometry helpers (matrix generation +
  validation, section geometry, touch-to-section mapping, navigation) with
  **no lvgl dependency**, so this logic is unit-testable on a desktop
  Python interpreter without a display or firmware build.
- `src/gui/screens/seedqr.py`: `SeedQROverviewScreen` + `SeedQRZoomScreen`,
  and the `show_seedqr()` overview<->zoom navigation loop. Both screens
  render the *same* canonical module matrix (generated once via
  `seedqr.generate_matrix()`), so the overview and every zoomed section
  can never disagree about module data.
- `src/keystore/ram.py`: routes the Standard/Compact SeedQR menu choices
  to `show_seedqr()`; the Plaintext QR path is unchanged except it now
  passes `sensitive=True`.
- Section sizing matches SeedSigner: a 21x21 QR groups into 7x7-module
  sections (3x3 grid); every larger SeedQR size groups into 5x5-module
  sections. A non-square-divisible size (e.g. 29x29) has a smaller
  trailing row/column, which is rendered at its true, smaller size (not
  padded out with filler modules).

## Security
This handles wallet recovery material, so:
- The mnemonic, SeedQR digit string, Compact SeedQR bytes, and the QR
  module matrix are never printed, logged, or persisted.
- While reviewing `gui/components/qrcode.py` I found a pre-existing
  unconditional `print(text)` that logged the full QR payload **even on
  real hardware**, not just the simulator. Removed it, and added a
  `sensitive` flag (threaded through `add_qrcode`/`QRAlert`) so simulator
  debug logging can be suppressed per-QR; it's now set for every
  mnemonic-related QR, including the pre-existing Plaintext export path.
- Both new screens drop their references to the secret matrix on LVGL
  `DELETE`.

## Companion fix (separate repo)
While building the matrix-generation helper I found a pre-existing bug in
the `f469-disco` submodule's `qrcode` usermod: `_qrcode_encode()` decided
text-vs-binary mode by scanning for an embedded `0x00` byte, so a `bytes`
payload with no zero byte (common for Compact SeedQR's raw entropy) fell
through to `qrcodegen_encodeText()`, which expects a NUL-terminated string
and could read past the intended data into uninitialized stack memory.
Fixed at diybitcoinhardware/f469-disco#43 (using the actual Python object
type instead of content-sniffing). I don't have write access to that
repo's upstream, so it's a separate PR; the `f469-disco` submodule pointer
in this PR is **unchanged** (still upstream's commit) — once #43 merges,
the submodule pointer here should be bumped in a follow-up commit.

## Test plan
- [x] `python3 run_native_tests.py` — 41/41 tests pass (encoding, section
      geometry, touch mapping, navigation, and print/log security
      regressions for the pure `seedqr` module).
- [x] `python3 -m compileall src test` — clean.
- [x] `make unix` (real firmware build, including this submodule fix) —
      builds successfully; confirmed the compiled `qrcode.encode_to_string()`
      produces the correct matrix size for all 4 SeedQR combinations
      (12/24-word x Standard/Compact), including entropy with no embedded
      zero byte.
- [x] Manually exercised all 4 combinations in the unix simulator: full
      overview, tapping into first/middle/last/partial sections, all four
      nav directions, back-to-overview section 
- [ ] Not tested on physical hardware.

This PR implements the SeedQR transcription mode requested in #212, including the sectioned overview and enlarged grid-based navigation for copying the QR to paper or metal.

Fixes #212

Screenshot 2026-07-28 112219
Screenshot 2026-07-28 112239
Screenshot 2026-07-28 114854
Screenshot 2026-07-28 122510
Screenshot 2026-07-28 122524
Screenshot 2026-07-28 122503

### DIFF
--- a/src/gui/common.py
+++ b/src/gui/common.py
@@ -176,7 +176,7 @@ def align_button_pair(btn1, btn2):
     btn2.set_x(HOR_RES // 2 + PADDING // 2)
 
 
-def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
+def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None, sensitive=False):
     """Helper functions that creates a title-styled label"""
     if scr is None:
         scr = lv.scr_act()
@@ -185,6 +185,7 @@ def add_qrcode(text, y=QR_PADDING, scr=None, style=None, width=None):
         width = 350
 
     qr = QRCode(scr)
+    qr.sensitive = sensitive
     qr.set_text(text)
     qr.set_size(width)
     qr.set_text(text)
--- a/src/gui/components/qrcode.py
+++ b/src/gui/components/qrcode.py
@@ -38,6 +38,9 @@ class QRCode(lv.obj):
 
         self.encoder = None
         self._autoplay = True
+        # When True, this QR encodes secret material (mnemonic, SeedQR
+        # digits/bytes): never print its contents, even in the simulator.
+        self.sensitive = False
 
         self.qr = lvqr.QRCode(self)
         self._text = "Text"
@@ -244,7 +247,7 @@ class QRCode(lv.obj):
         self.check_controls()
 
     def set_text(self, text="Text", set_first_frame=False):
-        if platform.simulator and self._text != text:
+        if platform.simulator and not self.sensitive and self._text != text:
             print("QR on screen:", text)
         self.encoder = None
         self._text = text
@@ -288,8 +291,8 @@ class QRCode(lv.obj):
         self.play.set_hidden((not self.is_fullscreen) or (self.idx is not None) or (self.encoder is None))
 
     def _set_text(self, text):
-        # one bcur frame doesn't require checksum
-        print(text)
+        if platform.simulator and not self.sensitive:
+            print(text)
         self.set_style(qr_style)
         self.qr.set_text(text)
         self.qr.align(self, lv.ALIGN.CENTER, 0, -100 if self.is_fullscreen else 0)
--- a/src/gui/screens/__init__.py
+++ b/src/gui/screens/__init__.py
@@ -8,3 +8,4 @@ from .input import PinScreen, InputScreen, DerivationScreen, NumericScreen
 from .mnemonic import MnemonicScreen, NewMnemonicScreen, RecoverMnemonicScreen
 from .transaction import TransactionScreen
 from .settings import DevSettings
+from .seedqr import SeedQROverviewScreen, SeedQRZoomScreen, show_seedqr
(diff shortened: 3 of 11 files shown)


Schnuartz:
@al-munazzim could you review that?

al-munazzim (approved):
Thanks for the thorough implementation and test coverage. I reviewed the SeedQR viewer changes, including the pure `seedqr` helpers, the LVGL overview/zoom screens, the mnemonic export routing, and the QR logging/sensitive handling.

What I checked locally:

- `python3 -m compileall -q src test`
- `python test/run_native_tests.py` in a fresh uv venv with `test/integration/requirements.txt` installed: 41/41 tests passed
- Diff/read-through of the security-sensitive paths that handle mnemonic / Standard SeedQR / Compact SeedQR material

The design looks sound to me:

- Standard and Compact SeedQR now route to the transcription viewer without rendering the raw digit string / entropy bytes as text.
- The viewer uses one canonical QR module matrix for overview and zoom, so the two views cannot diverge.
- Section sizing and partial trailing sections are covered by tests.
- The previous QR payload logging issue is addressed with the `sensitive` flag and the plaintext mnemonic QR path is marked sensitive too.
- CI is green.

No blocking findings from my review. Approving.

Schnuartz:
Tested on Hardware and it seems that we are having RAM issues.
IMG_20260815_142019

I did some other stuff before. And then I got this. Also I saw that the device got slower over time. 

I then shut the device don and tried to start again. And after starting we had some similar probelms again this time it freezed. IMG_20260815_142440

Issue #370, opened 2026-08-01
rng: TRNG timeout returns zeros silently; no sanity check in src/rng.py


helmchen1968:
I reviewed the entropy path after the Coldcard Mk3 advisory (July 2026) to check whether Specter shares that failure class. **It does not** — and the design is notably more robust:

- `boards/STM32F469DISC/mpconfigboard.h` sets `MICROPY_HW_ENABLE_RNG (1)`, so the Yasmarang software fallback in `ports/stm32/rng.c` is compiled out entirely.
- There is no second `rng_get()` implementation competing for the same symbol, which was the root cause on Coldcard.
- If the macro were ever misconfigured, `os.urandom` would not exist at all (`moduos.c`), so `src/rng.py` would fail loudly instead of silently degrading.
- Touchscreen entropy is fed automatically on every `PRESSING` event via the decorators in `src/gui/decorators.py`, not as an opt-in step.

While reading, I noticed two related gaps that I think are worth closing.

## 1. `rng_get()` returns 0 on timeout

In `f469-disco/micropython/ports/stm32/rng.c`:

```c
while (!(RNG->SR & RNG_SR_DRDY)) {
    if (HAL_GetTick() - start >= RNG_TIMEOUT_MS) {
        return 0;
    }
}
```

If the RNG peripheral stops responding, `rng_get()` returns zero without signalling an error. `os_urandom()` then fills its buffer with those zeros, and `src/rng.py` receives them as if they were entropy.

This is upstream MicroPython behaviour rather than Specter's code, but it is on the seed-generation path.

## 2. No sanity check in `get_random_bytes()`

In `src/rng.py`, the output of `get_trng_bytes()` is used and fed into the pool without any check:

```python
def get_random_bytes(nbytes):
    global entropy_pool
    d = get_trng_bytes(nbytes)
    feed(d)  # why not?
    ...
```

Combined with (1), a dead or stalled peripheral is indistinguishable from a working one.

## Impact

Low, and I want to be clear about that. The entropy pool mixes in touchscreen timing (`time.ticks_cpu()` at 180 MHz), so a failed TRNG would not make seeds predictable — it would silently reduce them to whatever the touch entropy contributed. That is defence in depth working as intended.

But it is a side effect rather than a guaranteed property. `entropy_pool` starts from the constant `b"7" * 64`, so the residual strength depends entirely on how many touch events happened to be accumulated beforehand — which is not measured anywhere.

## Suggestion

[...]

PR #372, opened 2026-08-03
Reject dead-TRNG output in get_random_bytes (#370)

White7hat7:
Implements the sanity check proposed by @helmchen1968 in #370.

Adds an application-side sanity check on TRNG output in `src/rng.py`. `rng_get()` in the STM32 port returns `0` on timeout, so a stalled peripheral currently surfaces as all-zero bytes that are indistinguishable from entropy. This rejects that failure mode with a visible `RNGError` instead of continuing silently.

The check lives in `rng.py` rather than the vendored `ports/stm32/rng.c`, so it doesn't diverge from upstream MicroPython.

`_looks_dead()` skips buffers under 4 bytes, so the existing `get_random_bytes(1)` call in `input.py` is unaffected — a single repeated byte is only treated as failure in bulk output.

**Testing**
- Native test suite passes unchanged (`run_native_tests.py`: 5/5 ok) with the patch applied.
- Manual verification of `get_random_bytes`: healthy 32-byte output passes; `get_random_bytes(1)` (keypad path) works; all-zero and single-repeated-byte output raise `RNGError`.

**Context**
I built the firmware and checked the entropy path at the object level to confirm it isn't the Coldcard class: `rng_get` has a single definition across the build (`nm`), the disassembly reaches `RNG->DR` at `0x50060800`, and the Yasmarang symbols in `modurandom.o` have internal linkage.

**Not included here** (deliberately, to keep scope tight): seeding `entropy_pool` from the first valid TRNG read instead of the `b"7" * 64` constant. Happy to follow up in a separate PR if the approach is agreed.


al-munazzim (approved):
Thanks for the focused fix. I reviewed this against #370 and the changed entropy path.

What I checked:

- The PR only changes `src/rng.py`.
- The `_looks_dead()` threshold avoids the known `get_random_bytes(1)` keypad-shuffle caller while still catching the seed/secret/IV sizes that matter (`16+`, `32`, and user-requested random output).
- On detected dead output it raises before `feed(d)`, so all-zero/repeated TRNG bytes do not poison the entropy pool.
- False positives for normal crypto-sized requests are negligible; the only realistic non-negligible case is a 4-byte request, and I do not see a security-sensitive fixed 4-byte caller in the current tree.

Local verification:

```text
# Ran 5 tests: OK

python -m compileall src test
# completed; only a pre-existing SyntaxWarning in src/keystore/flash.py

CPython smoke checks for _looks_dead() / RNGError / get_random_bytes(1)
# passed
```

I also attempted the full simulator `make test` path after installing the workflow system deps (`libsdl2-dev`, etc.). The local unix MicroPython simulator currently segfaults when importing `rng` on both this PR and `origin/master`, so I do not consider that PR-induced.

No blocker from my side.

al-munazzim (changes requested):
Follow-up after another look: I missed that this PR adds security-sensitive RNG behavior without any committed regression test coverage.

The code itself still looks reasonable to me, but please add tests for the new behavior before merge. At minimum, cover:

- `_looks_dead()` returns false for buffers shorter than 4 bytes, including repeated bytes, so the existing `get_random_bytes(1)` keypad-shuffle caller remains safe.
- `_looks_dead()` returns true for all-zero and single-repeated-byte buffers of length >= 4.
- `get_random_bytes()` raises `RNGError` before feeding the entropy pool when `get_trng_bytes()` returns dead output.
- A normal/non-repeated TRNG buffer still returns the expected number of bytes.

I previously ran these as local smoke checks, but they are not present in the repository test suite, and this path is important enough that the regression should live in CI.

al-munazzim:
I opened an alternative PR that keeps this fix and adds the missing RNG regression tests: #374. If preferred, the test commit can also be cherry-picked into this branch.

k9ert (commented):
Reviewed. I've cherry-picked the regression tests from #374 (8f7f8f2, authored by @al-munazzim) onto this branch, so this PR now carries both the fix and its tests. Verified locally: `python test/run_native_tests.py` → 11 tests, OK.

## Verdict

Sound change, correctly placed. Merge-worthy after two fixes. The core logic — check `d` *before* `feed()`, skip buffers <4 bytes so the `get_random_bytes(1)` keypad shuffle keeps working — is right, and the cherry-picked test asserting `entropy_pool` is unchanged after `RNGError` pins exactly the property that matters.

Worth stating explicitly, because it's the strongest argument for this PR: for `nbytes <= 64` the return value is already `sha512(pool || d)`, so a dead TRNG is *mostly* masked by accumulated touch entropy from `feed_touch()`. The check earns its keep in the one genuinely catastrophic case — dead TRNG at fresh boot, when `entropy_pool` is still the constant `b"7" * 64`. There the generated seed would be deterministic and **identical across all devices**. That's the scenario this closes, and it's worth a line in the commit message.

## Should fix

**1. Intermittent failure slips through — `src/rng.py:34`**

`len(set(data)) <= 1` only catches *total* failure. If `rng_get()` times out intermittently rather than permanently, you get e.g. 31 zeros + 1 random byte → `len(set(...)) == 2` → passes, with ~8 bits of real entropy in a 32-byte seed. That's the same silent-degradation class #370 is about.

**Correction to my original snippet:** I first suggested a bare `len(set(data)) < len(data) // 4`. That is wrong for large buffers — distinct-byte count saturates at 256, so a linearly growing threshold overtakes it. Measured false-positive rate on healthy `urandom`: **25% at n=1000, 99% at n=1024**. Since `getrandom` accepts up to 1000 bytes, that would have been a live bug. Capping the threshold fixes it:

```python
def _looks_dead(data):
    n = len(data)
    if n < 4:
        return False
    distinct = len(set(data))
    if n < 16:
        return distinct <= 1
    return distinct < min(n // 4, 32)
```

Measured (20k trials per size): **0 false positives** at n = 4, 8, 16, 32, 64, 128, 256, 512, 1000. Detection on a 32-byte seed made of `k` live bytes plus stuck zeros: 100% caught for k ≤ 6 (≤48 bits), tailing off at k = 7. `get_random_bytes(1)` keypad path unaffected.

To be clear about what this buys: it is still a liveness check, not a health check — it cannot see bias or a stuck LFSR with normal byte diversity. It moves the detection floor from "only catches a fully dead peripheral" to "catches partial stalls leaving under ~50 bits". Your docstring's caveat stays accurate.

**2. `RNGError` should subclass `BaseError` — `src/rng.py:16`**

`specter.py:140` catches `BaseError` and shows a clean alert using `e.NAME`. `specter.py:146` catches everything else and renders `"Something unexpected happened..."` plus a raw Python traceback. As written, a detected hardware RNG failure — the most security-critical error this device can surface — shows up as an anonymous crash dump.

Same in `cross_app_communicate` (`specter.py:716`): a `BaseError` becomes a meaningful `HostError` sent to the host, a plain `Exception` becomes a traceback.

```python
from errors import BaseError

class RNGError(BaseError):
    NAME = "RNG Error"
```

## Minor

- **Black formatting.** The blank line after `except:` at `src/rng.py:11` was removed; `master` is black-clean, this branch isn't (`black --check src/rng.py` fails). CI only runs `run_native_tests.py`, so nothing catches it.
- **Test gap: the `> 64` branch.** `src/rng.py:45` returns raw `d` with no pool mixing — the one path where `_looks_dead` is the *only* line of defence. Currently untested; please add a `get_random_bytes(100)` case.
- **Success path under-asserts.** The live-output test checks length only. Asserting `entropy_pool` changed would keep a dropped `feed()` call from regressing silently.
- **Dropped comment.** The original candid note (`"probably not the best way at the moment, but anything is better than nothing"`) was deleted, leaving `"assuming that entropy_pool has some real entropy"` dangling without its caveat. Unrelated to this fix — please restore it.
- **`getrandom` app.** `src/apps/getrandom.py:42` takes a host-chosen `num_bytes`; a 4-byte request has a 2^-24 chance of a spurious `RNGError` on healthy hardware. Acceptable as-is, and the threshold in fix 1 sidesteps it anyway.
- **Nit:** `data.count(data[0]) == len(data)` avoids the set allocation — cheaper on embedded than `len(set(data))`.

One thing I could not verify: the docstring's claim that `rng_get()` returns 0 on timeout. The `f469-disco` submodule isn't checked out in my working tree, so I couldn't read `ports/stm32/rng.c` directly. It matches upstream MicroPython and #370 quotes the code, so I have no reason to doubt it — just flagging it as unconfirmed on my end rather than checked.

k9ert (approved):

al-munazzim (approved):
Re-reviewed the latest commit eaff08125b294e24940cb98953fa5f61b0dffd4a.\n\nThe requested regression coverage is now committed in test/tests_native/test_rng.py and covers the short repeated buffers, repeated/dead output from 4 bytes onward, error-before-feed behavior, live output, partial stalls, the >64-byte raw TRNG path, and the long-buffer threshold cap. The RNGError now subclasses BaseError with a NAME, so the existing UI/host error handling paths surface a readable error instead of treating this as unexpected.\n\nI also ran the native suite locally in a disposable uv venv after installing test/integration/requirements.txt:\n\n    python test/run_native_tests.py\n\nResult: 16 tests passed. CI is green as well. Looks good to me now.

src/rng.py
```diff
@@ -14,16 +15,59 @@ def get_trng_bytes(nbytes):
...
+    n = len(data)
+    if n < 4:
+        return False
+    # iterating bytes yields one int per byte, so set() collapses the buffer
+    # to the distinct byte values it contains and len() counts them:
+    # b"\x00\x00\x00\x00" -> {0} -> 1, bytes(range(8)) -> {0..7} -> 8
+    distinct = len(set(data))
+    if n < 16:
+        # too short for a meaningful distribution - only reject a dead flatline
+        return distinct <= 1
```
maggo83: 16 Bytes is the length of a 12 word seedphrase's emtropy

even in 16 bytes we should way more than just 2 different byte values to assume this is healthy entropy.

src/rng.py
```diff
@@ -14,16 +15,59 @@ def get_trng_bytes(nbytes):
...
+    # a healthy TRNG gives ~30 distinct values in 32 bytes, so a buffer far
+    # below that is broken hardware, not bad luck: P(<=8 distinct in 32
+    # random bytes) is about 2^-121.
+    #
+    # The cap matters: there are only 256 possible byte values, so distinct
+    # saturates near 256 for long buffers while n // 4 keeps growing. Without
+    # min(..., 32) the check would reject ~25% of healthy 1000-byte requests
+    # (getrandom allows up to 1000) and almost every request above that.
+    return distinct < min(n // 4, 32)
```
maggo83: if distinct saturates near 256, why cap it here already at 32?

src/rng.py
k9ert: Agreed, and the concern is sharper than the anchor line suggests: 16 doesn't hit the `n < 16` branch, it hits `distinct < min(16 // 4, 32)`, so a 12-word seed only needed **4** distinct values. Healthy 16-byte output gives ~15.5.

`helpers.py:25` does `mnemonic_from_bytes(get_random_bytes(num_words * 4 // 3))`, so 12 words is exactly 16 bytes and 24 words exactly 32 — both were sitting on the loosest part of the curve.

The headroom was never used. Exact false-reject probabilities for the old thresholds:

| n | E[distinct] | old threshold | P(false reject) |
|---|---|---|---|
| 16 | 15.5 | <4 | 2^-69 |
| 32 | 30.1 | <8 | 2^-112 |
| 128 | 100.9 | <32 | 2^-249 |

Fixed in 29b6185: the threshold is now half of `256 * (1 - (255/256)**n)`, which puts 16 bytes at "reject below 8 distinct" (false reject 2^-40) and 32 bytes at "reject below 15" (2^-61).

src/rng.py
k9ert: Right — a cap is needed (`n // 4` outgrows the 256-value ceiling, that part of the reasoning holds), but 32 was arbitrary and much too low. At n=128 healthy output gives ~101 distinct and the cap accepted 32; at n=1000 it accepted 32 against an expectation of 251.

The cap is gone in 29b6185. The threshold now scales with the expectation itself — `2 * distinct < 256 * (1 - (255/256)**n)` — so it tracks the saturation curve instead of guessing where to stop growing.

`_expected_distinct()` is computed in Q32 fixed point rather than floats, so the threshold is identical on the F469 (single precision) and in CPython tests (double). Verified equal to `floor(256 * (1 - (255/256)**n))` for every n in 1..1000.

k9ert:
Reviewed and pushed a fix to the branch (29b6185) — @White7hat7 shout if you'd rather have taken it yourself, happy to drop it.

@maggo83's two points were both right, and chasing them turned up a third: **counting distinct values is the wrong statistic for the failure mode the docstring names.** It claims to catch a peripheral that "stalls only intermittently … mostly-repeated output with a few live bytes mixed in", but that's exactly what slipped through:

- 32 bytes, 25 zeros + 7 live → `distinct = 8`, threshold `< 8` → **passed**
- 16 bytes (a 12-word seed), 13 zeros + 3 live → `distinct = 4`, threshold `< 4` → **passed**, i.e. ~24 bits of real TRNG behind a 12-word mnemonic

So `_looks_dead()` now runs two checks:

1. **No byte value covers more than half the buffer** (n >= 8). This is what catches a partial stall. Gated at 8 because `b"\x00\x00\x00\x01"` is legitimate healthy output at n=4.
2. **Distinct values >= half of `256 * (1 - (255/256)**n)`**, replacing `min(n // 4, 32)`. Tracks the saturation curve rather than a magic cap.

`_expected_distinct()` uses Q32 fixed point, not floats, so the threshold doesn't shift between single-precision MicroPython on the F469 and double-precision CPython in the tests.

**False-rejection budget** (exact, not estimated):

| n | dominance check | count check |
|---|---|---|
| 4 | n/a | 2^-24 |
| 8 | 2^-26 | 2^-30 |
| 16 | 2^-51 | 2^-40 |
| 32 | 2^-99 | 2^-61 |
| 128 | 2^-388 | 2^-135 |

Worst case is 2^-24 at 4 bytes, which is unchanged from the current `distinct <= 1` behaviour and has no in-tree caller anyway (callers are 1, 16, 32, and <=1000 via `getrandom.py`).

Tests: 20/20 native pass, including 700 real `urandom` buffers across 7 sizes with no false rejects, both stall shapes, and `_expected_distinct` checked against the closed form.

Two notes for the record:

- The inline comment claiming `P(<=8 distinct in 32) ~ 2^-121` was off; exact value is 2^-111.6. Corrected.
- #374 is closed and its tests are already here (ad56f13), so nothing outstanding from that thread.

Placement is right, by the way — app-side in `rng.py` rather than patching the vendored `ports/stm32/rng.c`. And `RNGError(BaseError)` means `specter.py`'s handler shows a readable alert instead of a traceback. This was threshold tuning, not a redesign.

maggo83:
Is someone somewhere already working on a fix in the actual micropython code? Cause that random generator should never return just numerical 0 in case of a failure.
If not: I can raise it. But in my opinion we should also have a vendored solution for that until it is fixed upstream

src/rng.py
```diff
@@ -14,16 +15,90 @@ def get_trng_bytes(nbytes):
...
+
+    The check is skipped for very small requests, where a repeated byte can
+    occur legitimately and is not evidence of failure.
+    """
+    n = len(data)
+    if n < 4:
+        return False
+    # iterating bytes yields one int per byte, so counts maps byte value ->
+    # occurrences and top is the largest of those counts
+    counts = {}
```
maggo83, line 80: Directly from Kimi, not sure how relevant:

_looks_dead allocates a dictionary on every checked call — it builds up to a 256-entry counts dict each invocation, including for requests over 64 bytes that return raw TRNG output. On the constrained F469 MicroPython heap that adds GC pressure and non-deterministic latency on the seed/key-generation path; a fixed 256-element counter array would be cheaper and allocation-free.

test/tests_native/test_rng.py
```diff
@@ -0,0 +1,123 @@
...
+
+    setup_native_stubs()
+
+from unittest import TestCase
+
+import rng
+from errors import BaseError
+
+
+class RNGSanityCheckTest(TestCase):
```
maggo83, line 15: Maybe add a test case for 0 length as a special/edge case:

get_random_bytes(0) would call get_trng_bytes(0), _looks_dead(b"") returns False, then feed(b"")

test/tests_native/test_rng.py
```diff
@@ -0,0 +1,123 @@
...
+        data = bytes(range(245)) * 4 + bytes(range(20))
+        self.assertEqual(len(data), 1000)
+        self.assertEqual(len(set(data)), 245)
+        self.assertFalse(rng._looks_dead(data))
+
+    def test_looks_dead_passes_real_random_output(self):
+        # guards against a threshold tight enough to reject healthy hardware
+        for nbytes in (4, 8, 16, 32, 64, 128, 1000):
+            for _ in range(100):
+                self.assertFalse(rng._looks_dead(os.urandom(nbytes)))
```
maggo83: Statistical os.urandom test can flake in theory. 700 samples with a claimed false-rejection bound below 2^-24 for the smallest size gives a per-run flake probability around 4e-5 — low, but not zero across many CI runs.

A fixed seed would remove that.

test/tests_native/test_rng.py
k9ert: Good catch, and it turns out to be more than an edge case: `get_random_bytes(0)` is reachable from the host. `apps/getrandom.py` rejects `num_bytes < 0` and caps at 1000, but `0` passes straight through.

The path behaves as you describe — `get_trng_bytes(0)` → `b""`, `_looks_dead(b"")` returns `False` (the `n < 4` guard), `feed(b"")` still advances the pool, and the caller gets `b""` rather than an error. Nothing was pinning that down.

Covered end to end in fb4b5ee, including the assertion that the pool is still advanced.

test/tests_native/test_rng.py
k9ert: Agreed in principle — fixed in fb4b5ee. The magnitude is smaller than the estimate, but not zero, and removing it is free.

Exact false-rejection probabilities at the current thresholds (full enumeration over count-partitions, not simulated):

| n | reject condition | P(false reject) |
|---|---|---|
| 4 | distinct ≤ 1 | 5.96e-08 |
| 8 | top ≥ 5, or distinct ≤ 3 | 1.36e-08 |
| 16 | top ≥ 9, or distinct ≤ 7 | 6.43e-13 |

100 samples × 7 sizes gives **~7.3e-6 per run**, so roughly 1 flake in 137k CI runs rather than the ~4e-5 estimated. Everything from 32 bytes up is negligible; the 4-byte case dominates.

`os.urandom` can not be seeded, but `tests_native` only runs under CPython, so the sweep now draws from `random.Random(0x5EEDBEEF)`. Uniform independent bytes are exactly what the thresholds are specified against, so the test keeps its purpose — guarding against a threshold tight enough to reject healthy hardware — without the dice roll. Confirmed it still fails when the threshold is tightened from half the expectation to 0.95 of it.

src/rng.py
k9ert: Thanks — I looked at this one closely and I would rather not make the change, because I think it is a pessimization for the sizes that actually run on the device.

The premise is that the dict grows "up to 256 entries". It is bounded by the *distinct byte values in the buffer*, not by 256. Actual call sites:

| caller | n | dict entries |
|---|---|---|
| `input.py:233` keypad shuffle | 1 | **0** — `n < 4` returns before the loop |
| `helpers.py:41` AES IV (every encrypt) | 16 | ≤ 16 |
| `helpers.py:25` 12-word seed | 16 | ≤ 16 |
| keystore secrets / 24-word seed / default `getrandom` | 32 | ≤ 32 |
| `apps/getrandom.py` host request | ≤ 1000 | ≤ 256 |

A fixed 256-element counter array allocates 8–16× more than the dict it replaces on every path that runs on the device, and only pays off for the rare host-driven 1000-byte request. The seed and key-generation path the comment is concerned about is 16 or 32 bytes — roughly 32 dict operations and a dict that never exceeds 32 slots. And the hot path, the keypad, allocates nothing today because of the `n < 4` early return.

If allocation-free were genuinely worth chasing, the right shape would be `set(data)` plus `bytes.count` per distinct value — both C-level, one small set — rather than a 256-slot array. But that depends on `MICROPY_PY_BUILTINS_STR_COUNT` in the f469-disco port, and the win over the current code is on the order of tens of dict operations, so it is not worth the added coupling.

Happy to revisit if someone measures actual GC pressure on hardware, but I do not think the change stands on the reasoning as given.

maggo83 (approved):

schnuartz-ai:
One more point on @maggo83's upstream question: upstream MicroPython has already fixed exactly this — micropython/micropython@240240e ("stm32/rng: Raise an exception if hardware RNG fails", committed Aug 14) raises `OSError("RNG failed")` on TRNG timeout instead of returning 0.

Our pinned fork still has the old behavior: specter-diy pins f469-disco @ db3ce3e → diybitcoinhardware/micropython @ 6bdf1b6, whose `ports/stm32/rng.c` still does `return 0;` on timeout. I'd suggest backporting the upstream fix to the fork as defense-in-depth (PR against diybitcoinhardware/micropython, then bump the submodule chain). The fork's rng.c even predates the CONDRST retry logic, but that retry does not apply to the F469 anyway (the F4 RNG has no CONDRST bit), so a minimal backport — raise instead of `return 0;` — is exactly equivalent to upstream behavior on this hardware.

The app-side check in this PR stays necessary either way: the low-level fix only catches a full timeout, while `_looks_dead()` also catches partial stalls.

One thing worth doing right here: once the low-level RNG raises, `get_trng_bytes()` (= `os.urandom`) will throw OSError, which specter.py would render as "Something unexpected happened" plus a traceback. I've posted a small translation to RNGError in rng.py plus a regression test as one-click-appliable suggestions below, so the failure still surfaces through the clean "RNG Error" alert path this PR introduced. It is a no-op until the fork is updated.

src/rng.py
```diff
@@ -14,16 +15,90 @@ def get_trng_bytes(nbytes):
...
+    if n >= 8 and top * 2 > n:
+        return True
+    return 2 * len(counts) < _expected_distinct(n)
+
+
+# assuming that entropy_pool has some real entropy
+# we can generate bytes using it as well
 def get_random_bytes(nbytes):
     global entropy_pool
     d = get_trng_bytes(nbytes)
```
schnuartz-ai: Defense-in-depth companion to the low-level fix from the thread: once the STM32 port raises `OSError("RNG failed")` on TRNG timeout (upstream micropython/micropython@240240e, proposed for backport to our fork), `get_trng_bytes()` will throw and specter.py would show "Something unexpected happened" plus a traceback. Translating it here keeps the failure on the readable `RNGError` alert path. No-op until the fork actually raises.

```suggestion
    try:
        d = get_trng_bytes(nbytes)
    except OSError:
        # upstream micropython@240240e raises OSError("RNG failed") on
        # TRNG timeout; our fork still returns 0 until it is updated
        raise RNGError("TRNG hardware failure")
```

test/tests_native/test_rng.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_get_random_bytes_returns_raw_trng_above_64_bytes(self):
+        rng.get_trng_bytes = lambda nbytes: bytes(range(nbytes))
+        self.assertEqual(rng.get_random_bytes(100), bytes(range(100)))
+
+    def test_rng_error_is_a_base_error(self):
+        # BaseError subclasses get a readable alert in specter.py instead of
+        # an "unexpected error" traceback
+        self.assertTrue(issubclass(rng.RNGError, BaseError))
+        self.assertEqual(rng.RNGError.NAME, "RNG Error")
```
schnuartz-ai: Regression coverage for the OSError → RNGError translation suggested on `src/rng.py`: low-level raise must surface as RNGError, and the entropy pool must stay untouched.

```suggestion
        self.assertEqual(rng.RNGError.NAME, "RNG Error")

    def test_get_random_bytes_translates_oserror_to_rng_error(self):
        # once the STM32 port raises on TRNG timeout instead of returning 0
        # (micropython/micropython@240240e), the user should still get the
        # readable RNGError alert rather than an unexpected-error traceback
        rng.entropy_pool = b"A" * 64

        def failing_trng(nbytes):
            raise OSError("RNG failed")

        rng.get_trng_bytes = failing_trng
        try:
            rng.get_random_bytes(32)
        except rng.RNGError:
            pass
        else:
            self.fail("Expected RNGError when the TRNG raises OSError")
        self.assertEqual(rng.entropy_pool, b"A" * 64)
```

Schnuartz (approved):
Checked it and it is ready to merge from my pov. It makes the existing codebase more robust.

schnuartz-ai (approved):

PR #374, opened 2026-08-07
Reject dead TRNG output with RNG regression tests

al-munazzim:
This is an alternative to #372 that includes the same RNG sanity-check fix plus regression tests for the new behavior.

It keeps the application-side check proposed in #372:

- reject all-zero / single-repeated-byte TRNG output for buffers of length >= 4
- keep sub-4-byte requests working, especially the existing `get_random_bytes(1)` keypad-shuffle caller
- raise before feeding dead TRNG output into the entropy pool

Additional test coverage added here:

- `_looks_dead()` ignores short repeated buffers
- `_looks_dead()` rejects repeated buffers from 4 bytes upward
- `_looks_dead()` allows non-repeated buffers
- `get_random_bytes()` raises `RNGError` before changing `entropy_pool` on dead TRNG output
- `get_random_bytes(1)` still works with repeated one-byte output
- normal/non-repeated TRNG output returns the requested length

Verification run locally:

```text
# Ran 11 tests: OK

python -m compileall src test
# OK
```

Context: #370, #372.

### DIFF
--- a/src/rng.py
+++ b/src/rng.py
@@ -8,22 +8,41 @@ entropy_pool = b"7" * 64
 try:
     from os import urandom as get_trng_bytes
 except:
-
     def get_trng_bytes(nbytes):
         with open("/dev/urandom", "rb") as f:
             return f.read(nbytes)
 
 
+class RNGError(Exception):
+    """Raised when the hardware TRNG output fails a basic sanity check."""
+
+
+def _looks_dead(data):
+    """Detect a stalled or failed TRNG.
+
+    rng_get() in the STM32 port returns 0 on timeout (ports/stm32/rng.c),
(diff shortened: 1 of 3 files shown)


k9ert:
Thanks @al-munazzim — the regression tests here are the valuable part, and I'd like to keep them.

Since #372 came first and carries an identical `src/rng.py`, let's continue there rather than maintain two PRs for the same fix. I've cherry-picked your test commit (8f7f8f2) onto `White7hat7:fix/rng-sanity-check`, authorship preserved, so #372 now has both the fix and your tests. `python test/run_native_tests.py` → 11 tests, OK.

Treating this one as superseded by #372 — review comments are over there, please follow up on them there.

PR #376, opened 2026-08-08
Update embit to v0.8.2

miketlk:
## Summary

Update Specter-DIY for embit’s new nested `src/embit` submodule layout.

- Fix firmware manifests and Python paths.
- Exclude CPython-only embit modules from frozen firmware.
- Initialize nested submodules during builds.
- Use the local embit submodule in tests.
- Enable recursive submodule checkout in CI.
- Keep mixed multipath and fixed-path BIP 389 descriptors compatible with the
  previous embit version.

## Validation

- `make unix`
- `make test` — 19 passed, including frozen-import checks
- Native tests — 7 passed
- `make disco`

## Changes in `embit`

- Align PSBT version parsing and serialization with BIP 174 and BIP 370 across `PSBT` and `PSBTView`.
- Add `PSBT_IN_TAP_KEY_SIG` support and miniscript boolean literals `0` and `1`.
- Make BIP39 phrase and entropy validation stricter.
- Fix descriptor parser offsets, descriptor/miniscript branch counting, and
  miniscript `t:` wrapper validation.
- Replace assertion-based input checks with explicit exceptions and make PSBT
  signing output deterministic.
- Rework secp256k1 backend discovery and fallbacks, remove bundled native
  libraries, and require Python 3.10+ for CPython users.

Specter-DIY includes a focused compatibility shim for [embit#136](https://github.com/diybitcoinhardware/embit/issues/136), allowing descriptors that mix multipath and fixed-path keys while retaining rejection of genuinely mismatched multipath branch counts.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
@@ -29,6 +31,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
+        with:
+          submodules: recursive
       - name: Install dependencies
         run: |
           sudo apt-get update
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ BOARD ?= STM32F469DISC
 FLAVOR ?= SPECTER
 USER_C_MODULES ?= ../../../usermods
 MPY_DIR ?= f469-disco/micropython
+EMBIT_INIT ?= f469-disco/libs/common/embit/src/embit/__init__.py
 ifeq ($(shell uname),Linux)
     MPY_CFLAGS ?= -Wno-dangling-pointer -Wno-enum-int-mismatch
 else
@@ -22,8 +23,11 @@ $(TARGET_DIR):
 $(MPY_DIR)/mpy-cross/Makefile:
 	git submodule update --init --recursive
 
+$(EMBIT_INIT): | $(MPY_DIR)/mpy-cross/Makefile
+	git submodule update --init --recursive
+
 # cross-compiler
-mpy-cross: $(TARGET_DIR) $(MPY_DIR)/mpy-cross/Makefile
+mpy-cross: $(TARGET_DIR) $(MPY_DIR)/mpy-cross/Makefile $(EMBIT_INIT)
 	@echo Building cross-compiler
 	make -C $(MPY_DIR)/mpy-cross \
         DEBUG=$(DEBUG) \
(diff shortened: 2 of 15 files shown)


Schnuartz (approved):
I checked that code with the two very best Cybersecurity AI Modells Kimi K3 and Qwen 3.8 max.

This looks very well implemted.

al-munazzim (approved):
Reviewed the embit v0.8.2 update, nested submodule layout, Makefile/manifest changes, test path updates, and the descriptor compatibility shim.

Verification performed:
- `python3 test/run_native_tests.py` in a fresh uv venv: 7 tests passed.
- `python3 -m compileall ../src ../test`: completed successfully (only the pre-existing `is ""` SyntaxWarning surfaced).
- Manual CPython smoke with native stubs for mixed multipath/fixed-path descriptors through `Wallet.parse()` and `LDescriptor.from_string()`, plus the unequal-branch negative case: passed.
- Re-checked GitHub checks on the current head: Build Firmware, native-tests, tests, and Netlify deploy preview are green.

The manifest inventory tests cover the new frozen-module selection, the path ordering now correctly prefers `embit/src`, and the compatibility shim stays scoped to the specific embit v0.8.2 branch-length regression while preserving the unequal-multipath failure case.

No blockers found.

Schnuartz:
I did some QR-Code testings.

Here I used the QR-Codes that were extra difficult edge cases we used to test M3Y Scanner compatibility and GM65 Compatibility.
https://github.com/cryptoadvance/specter-diy/pull/335

Also I tried a Bitcoin Keeper Multi Sig Descriptor.

Everything works.

k9ert (approved):

pythcoiner:
> Specter-DIY includes a focused compatibility shim for https://github.com/diybitcoinhardware/embit/issues/136

concept NACK, I dont see a good reason to accept a malformed descriptor

as a note mixed multipath length are explicitly rejected by bitcoin core parser:

- multi/sortedmulti/multi_a/sortedmulti_a — descriptor.cpp:2440-2451

```cpp
        // Make sure all vecs are of the same length, or exactly length 1
        // For length 1 vectors, clone key providers until vector is the same length
        for (auto& vec : providers) {
            if (vec.size() == 1) {
                for (size_t i = 1; i < max_providers_len; ++i) {
                    vec.emplace_back(vec.at(0)->Clone());
                }
            } else if (vec.size() != max_providers_len) {
                error = strprintf("multi(): Multipath derivation paths have mismatched lengths");
                return {};
            }
        }

```

- tr() leaf subscripts — descriptor.cpp:2586-2597

```cpp
        // Make sure all vecs are of the same length, or exactly length 1
        // For length 1 vectors, clone subdescs until vector is the same length
        for (auto& vec : subscripts) {
            if (vec.size() == 1) {
                for (size_t i = 1; i < max_providers_len; ++i) {
                    vec.emplace_back(vec.at(0)->Clone());
                }
            } else if (vec.size() != max_providers_len) {
                error = strprintf("tr(): Multipath subscripts have mismatched lengths");
                return {};
            }
        }
```

- tr() internal key vs. subscripts — descriptor.cpp:2599-2602
  
```cpp
        if (internal_keys.size() > 1 && internal_keys.size() != max_providers_len) {
            error = strprintf("tr(): Multipath internal key mismatches multipath subscripts lengths");
            return {};
        }

```
- musig() participants — descriptor.cpp:2142-2145:
  
```cpp
        } else if (max_multipath_len > 1) {
            if (!clone_providers(max_multipath_len)) {
                error = strprintf("musig(): Multipath derivation paths have mismatched lengths");
                return {};
            }
```

al-munazzim:
Thanks for checking this. I read the cited Core snippets slightly differently: Core rejects true mismatched multipath branch counts, but it explicitly treats a length-1 vector as broadcastable by cloning it up to `max_providers_len`:

```cpp
if (vec.size() == 1) {
    for (size_t i = 1; i < max_providers_len; ++i) {
        vec.emplace_back(vec.at(0)->Clone());
    }
} else if (vec.size() != max_providers_len) {
    ... mismatched lengths ...
}
```

So I agree we should not accept genuinely malformed descriptors where multiple multipath providers have unequal branch counts. The compatibility case I reviewed was the narrower Core-compatible one: one side has a single fixed path / single branch and another side has the multipath branch list, which Core normalizes by cloning the singleton.

For merge safety I think the important test coverage is exactly that boundary:

- accept fixed-path/singleton + multipath descriptors that Core would clone,
- reject unequal multipath branch counts where neither side is a singleton.

If the PR's shim is broader than that, I agree it should be tightened before merge; if it is only doing the singleton broadcast case, I would not call that malformed relative to Core's parser behavior.

pythcoiner:
> but it explicitly treats a length-1 vector as broadcastable by cloning it up to max_providers_len

Right, my note was about the mixed multipath length, and I think core is not enough strict when importing such descriptor, [see](https://github.com/bitcoin/bitcoin/issues/35985)

my rationale here is still that it is a malformed descriptor,and I still dont think the descriptor rules import must be relaxed: I have no knowledge of any wallet building this kind of descriptor, nor user complaining about it, so imho it can only hurt users

PR #377, opened 2026-08-11
fix: don't let delete_mnemonic report success when deletion failed

schnuartz-ai:
## Summary

Security hardening: both `FlashKeyStore.delete_mnemonic()` and `SDKeyStore.delete_mnemonic()` had their `return True` inside the `finally` block of the delete path. A `return` inside `finally` executes *while* the `KeyStoreError` raised in the `except` block is still propagating — and silently discards it.

**Consequence:** when a deletion failed (I/O error, card removed mid-operation, read-only filesystem, …), the error never reached the caller, and the UI showed *"Your key is deleted."* while the key file was still sitting on the internal flash or the SD card — the exact opposite of what a user needs to hear when they are trying to get rid of key material.

## Fix

Move the success `return True` out of the `finally` block in both classes:

- `FlashKeyStore.delete_mnemonic()` (internal flash): drop the `finally` entirely, return after the try/except.
- `SDKeyStore.delete_mnemonic()` (flash + SD): the SD unmount stays in `finally` (it must run on every path); only the success return moves out.

On failure the `KeyStoreError` now propagates and is shown as an error (caught by `Specter.handle_exception` → `gui.alert`); on success nothing changes.

## Tests

New `test/tests_native/test_delete_mnemonic.py`, for both keystores:

- **failure path**: patched `os.remove` raising → `KeyStoreError` reaches the caller, and the file is provably still present (fails against the old structure, where the exception was discarded and `True` returned)
- **success path**: returns `True`, file is gone
- **cancel path**: picker cancelled → returns `False`, file untouched

`python3 test/run_native_tests.py`: 11/11 pass.

## Notes

- Same pattern exists in the Schnuartz fork; submitted here per the disclosure note in that fork's PR.
- Reviewer notes: this is a minimal, behavior-preserving fix for the exception-swallowing `return` in `finally`. A pre-existing mount/unmount asymmetry on the `File not found.` path in `SDKeyStore.delete_mnemonic()` (mount happens before the `file_exists` check, but the raise bypasses the `finally` unmount) is out of scope and not changed here.

### DIFF
--- a/src/keystore/flash.py
+++ b/src/keystore/flash.py
@@ -302,8 +302,10 @@ class FlashKeyStore(RAMKeyStore):
         except Exception as e:
             print(e)
             raise KeyStoreError("Failed to delete file '%s'" % file)
-        finally:
-            return True
+        # NOTE: no `return` in a `finally` block here - a return inside
+        # finally executes while an exception is propagating and silently
+        # discards it, so a failed delete would still report success.
+        return True
 
     async def get_input(
             self,
--- a/src/keystore/sdcard.py
+++ b/src/keystore/sdcard.py
@@ -162,7 +162,12 @@ class SDKeyStore(FlashKeyStore):
         finally:
             if platform.sdcard.is_present and file.startswith(self.sdpath):
                 platform.sdcard.unmount()
-            return True
+        # NOTE: this return must stay OUTSIDE the finally block - a return
+        # inside finally executes while an exception is propagating and
+        # silently discards it, so a failed delete would still report
+        # success. The unmount above belongs to the finally (it must run
+        # on every path); the success return does not.
+        return True
 
     async def storage_menu(self):
         """Manage storage, return True if new key was loaded"""
--- a/test/tests_native/__init__.py
+++ b/test/tests_native/__init__.py
@@ -1,1 +1,2 @@
 from .test_wallet_manager_parsing import *
+from .test_delete_mnemonic import *
--- a/test/tests_native/test_delete_mnemonic.py
+++ b/test/tests_native/test_delete_mnemonic.py
@@ -1,0 +1,109 @@
+"""
+Regression tests: delete_mnemonic() must not swallow deletion failures.
+
+Both FlashKeyStore.delete_mnemonic() and SDKeyStore.delete_mnemonic() had
+their `return True` inside the `finally` block of the delete path. A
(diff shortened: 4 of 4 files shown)


Schnuartz:
@al-munazzim could you review this?

schnuartz-ai:
Cross-reference: #378 ports the full BitBox02 backup import/export + secure-delete feature set from the Schnuartz fork to upstream, and includes the same \
eturn in finally\ fix as this PR (adapted for \secure_delete_file\ instead of \os.remove\).

**Merge order:** see #378's description — either merge this PR (#377) first and rebase #378 on top, or merge #378 and close this one as superseded. The two PRs conflict on \delete_mnemonic()\ in both \lash.py\ and \sdcard.py\, so they should not be merged independently without rebasing.

schnuartz-ai:
Updated cross-reference: a second upstream PR has been created.

**#379** — \secure file deletion + SD card format + delete_mnemonic error propagation fix\. This is a small (~500 lines), focused PR that contains the same \
eturn in finally\ fix as this PR (#377), plus \platform.secure_delete_file()\, \platform.secure_delete_tree()\ (with DoS cap), and \SDCard.erase_and_format()\. It switches \delete_mnemonic()\ from \os.remove\ to \secure_delete_file\.

**#378** — the full BitBox02 import/export + SD card menu feature (~5300 lines). Contains everything in #379 plus the BitBox parser/writer/import/export on top.

**Three open PRs now touch \delete_mnemonic()\.** See #379's description for the full merge-order matrix (Options A/B/C) and the \do not merge independently without rebasing\ warning. Recommended: merge this PR (#377) first, then rebase #379 on top, then rebase #378 on top.

schnuartz-ai:
Closing as superseded by #379, which contains the same \
eturn in finally\ fix (adapted for \platform.secure_delete_file\ instead of \os.remove\) plus the new \secure_delete_file\/\secure_delete_tree\/\erase_and_format\ machinery that \delete_mnemonic()\ now uses.

Merge order is now simplified to:
1. Merge #379 (secure-delete + delete_mnemonic fix)
2. Merge #378 (BitBox feature, rebased on top of #379)

The regression tests from this PR (\	est_delete_mnemonic.py\) are included in #379, patched to mock \platform.secure_delete_file\ instead of \os.remove\.

Thanks to @3rdIteration and @k9ert for the review on the standalone fix.

PR #378, opened 2026-08-12
SD card: BitBox02 backup import/export + secure delete + delete_mnemonic error propagation fix

schnuartz-ai:
## Summary

BitBox02 microSD backup import/export + unified SD card menu, built on top of the secure-delete foundation from #379.

### What this PR adds

**BitBox02 microSD backup import** (`Key management` -> `Import recovery phrase` -> `BitBox microSD backup`):
- From-scratch, independent reimplementation of the BitBox02 backup wire format (`src/bitbox_backup.py`, `src/bitbox_sd.py`) — no code copied from bitbox02-firmware. Strict protobuf parser (rejects non-canonical varints, groups, duplicate fields, unknown backup versions), checksum + backup-id verification, 3-copy majority-vote recovery.
- The BitBox microSD format is **not encrypted** and **not authenticated**. The import shows an explicit warning and never auto-saves the imported key.

**BitBox02-format export** (`Show recovery phrase` -> `Save to SD card`):
- Writes a real BitBox02-compatible backup (3 redundant copies under `bitbox02/<backup id>/`), verified by reading back through the same parser the import uses.
- Also offers plain-text (Specter format) and encrypted (this-device-only) options, each with honest warnings.

**Unified SD card menu** (`Device settings` -> `SD card`):
- Per-item secure delete (encrypted key files, plaintext exports, BitBox backup dirs) using `platform.secure_delete_file`/`secure_delete_tree` from #379.
- Full-card format via `SDCard.erase_and_format()` from #379.

## Dependency & merge order

**This PR (#378) depends on #379 and must be merged AFTER it.**

#378 is already rebased on top of #379: the first commit on this branch is identical to #379's commit. After #379 merges, click "Update branch" on this PR — GitHub will rebase cleanly (the shared commit falls away since it's already in master, leaving only the BitBox feature commit).

**Order:**
1. Merge #379 (secure-delete foundation)
2. Merge #378 (this PR — BitBox feature on top)

**Do not merge in the other order.** This PR uses `secure_delete_file` / `erase_and_format` / `secure_delete_tree` introduced in #379.

(#377 was closed as superseded by #379.)

## Test plan

- [x] `python3 -m compileall src test`
- [x] `python3 test/run_native_tests.py` (173/173 pass, includes BitBox parser/writer/flow, platform SD, delete menu, label validation, event-loop yield, zeroization)
- [ ] Manual click-through on device/simulator

## Notes

[...]

PR #345, opened 2026-02-19
Documentation: Add link to udev/README.md to mkdocs

maggo83:
Because the link is missing in mkdocs, the link to the udev rules in the main README.md breaks when accessing the readme via https://docs.specter.solutions/diy/

PR #379, opened 2026-08-12
feat: secure file deletion + SD card format + delete_mnemonic error propagation fix

schnuartz-ai:
## Summary

Small, focused security-hardening PR: secure file deletion, full-card SD format, and the `delete_mnemonic` error-propagation fix.

### What this PR adds

**1. `platform.secure_delete_file()` / `platform.secure_delete_tree()`**

Overwrites a file with zeros and flushes that overwrite to the device before unlinking it. A plain `os.remove()` only removes the directory entry — the file's old bytes remain physically on the card and can be recovered with an undelete tool until that space is reused. This closes that gap.

One pass of zeros, deliberately. NIST SP 800-88 asks for user-addressable data to be replaced with non-sensitive data, not for that data to be random, and multi-pass overwriting is a magnetic-media practice. Zeros also keep this off the hardware RNG — `secure_delete_tree()` is synchronous, so at its 8 MiB cap random data would block the firmware for over a minute.

`secure_delete_tree()` applies the same to every file in a directory tree. The tree is enumerated and size-checked **before any overwrite**, aborting mid-enumeration once a cap is exceeded (`SECURE_DELETE_MAX_ENTRIES`, `_MAX_DEPTH`, `_MAX_FILE_BYTES`, `_MAX_TOTAL_BYTES`) — so an adversarial directory cannot force unbounded traversal or overwrite work, and a partial wipe is never left behind. Traversal is iterative, not recursive, so a deep tree cannot exhaust the device's small fixed Python stack.

**2. `platform.SDCard.erase_and_format()`**

Overwrites **every block** on the SD card with zeros, then creates a fresh, empty FAT filesystem. Zeros rather than random data because `os.urandom()` on the pinned MicroPython calls `rng_get()` once per byte and each of those busy-waits for `RNG_SR_DRDY` (~10 µs by its own comment) — roughly three hours per gigabyte of RNG wait before a single SD write. NIST SP 800-88 asks for user-addressable data to be replaced with non-sensitive data, not for it to be random. Validates block geometry before starting, re-raises device errors as `RuntimeError` with actionable recovery text, and yields to the asyncio event loop after every chunk — also when a `progress_cb` is given, since a callback that never awaits would otherwise starve the GUI task for the whole erase. Cancellation at an await point is reported as an interrupted wipe rather than silently returning.

**3. `delete_mnemonic()` error propagation fix**

Both `FlashKeyStore.delete_mnemonic()` and `SDKeyStore.delete_mnemonic()` had `return True` inside the `finally` block, silently discarding the `KeyStoreError` raised when deletion failed. The UI showed *"Your key is deleted."* while the key file was still on disk. The success return is moved out of `finally`; both keystores now use `secure_delete_file()` instead of `os.remove()`.

**4. Shared geometry validation**

`is_valid_count()` / `validate_block_geometry()` are used by the SD erase, by the tree's file-size check, and by `platform.wipe()` — which previously validated no flash geometry at all. The helper also enforces a maximum block size and a `min_blocks` floor, so a hardcoded block range (as `wipe()` uses) can never run past the geometry the device reports. `is_block_op_success()` is the single place that knows which return values mean success — `pyb.SDCard.writeblocks()` returns `True`/`False`, `pyb.Flash.writeblocks()` an integer status, generic block devices `None`.

**6. `wipe()` now flushes before resetting**

`pyb.Flash` writes into a RAM cache that is only written back on an explicit `BDEV_IOCTL_SYNC` or after 5 seconds of idle (`flashbdev.c`), and `pyb.hard_reset()` reaches `NVIC_SystemReset()` without calling `storage_flush()`. `wipe()` wrote ~200 blocks and reset immediately, so the overwrite could be discarded with the cache and the device could come back up with the data still present. It now issues `ioctl(3)` after the last write, and checks every write status — `pyb.Flash.writeblocks()` returns its status rather than raising, so failures were silent. A failed write or flush raises rather than rebooting, and `Specter.wipe_or_report()` turns that into an explicit "the device was NOT wiped, your data may still be present" screen instead of an unhandled error; on the critical-error path a failed wipe keeps offering the wipe rather than continuing.

This bug predates the PR. `wipe()` is touched here only because this branch already added geometry validation to it.

**5. Bounded wipe allocations**

The erase buffer is allocated once, before the first destructive write, stays zero-filled and is never modified — per-chunk allocation risked a `MemoryError` partway through a wipe. A failure at that single allocation reports that nothing was changed; a `MemoryError` later is reported as an interrupted wipe rather than escaping raw. Cleanup in `finally` never replaces the error already propagating.

## Review status

Copilot's findings (geometry validation, ioctl error translation, cancellation reporting, short-write and strict-sync handling, cap enforced during enumeration rather than after) were addressed in 25746f4 / 55b9bb2 / 32c43f8.

@maggo83's review is addressed in 938f96b: single pass, iterative traversal, caps moved out of `secure_delete_file()` into `secure_delete_tree()`, shared sanity-check helpers, `progress_cb` failures isolated from the event-loop yield, named `has_block_device` property, and docstrings trimmed of cross-references and user-facing wording. Each thread has an inline reply.

## Merge order

**This PR (#379) must be merged FIRST.** Then merge #378 (BitBox02 import/export feature).

#378 is already rebased on top of this PR — after merging #379, clicking "Update branch" on #378 will cleanly rebase it (the secure-delete commit falls away since it's already in master, leaving only the BitBox feature commit).

**Do not merge in the other order.** #378 depends on `secure_delete_file` / `erase_and_format` / `secure_delete_tree` which are introduced here.

(#377 was closed as superseded — this PR contains its fix.)

## Test plan

- [x] `python3 -m compileall src test`
- [x] `python3 test/run_native_tests.py` — `Ran 64 tests ... OK`. Regression coverage:
  - `test_delete_mnemonic.py`: failure/success/cancel paths for both keystores, SD unmount cleanup, unmount failure not hiding a delete failure
  - `test_platform_sd.py`: `wipe()` flushes the flash cache after the last write and before reboot, writes the whole documented block range, aborts without rebooting on a failed write or failed sync, and can be retried after a failure (the stub raises `EINVAL` for an already-unmounted path, as MicroPython does); `erase_and_format` block coverage / event-loop yield / write-error / invalid geometry / interrupted wipe / power-on failure reported as a `False` return rather than an exception; full erase writes zeros and makes no `os.urandom()` calls at all; write buffer allocated exactly once; allocation failure before any write, and `MemoryError` mid-wipe, both reported as such; cleanup failure neither masking a wipe error nor swallowed on success; oversized block size and `min_blocks` floor; `mkfs()` called exactly once on the same device only after a complete overwrite, never after a failed one, and its failure translated; `unmount()` retry path keeps the card powered while `/sd` is still mounted; single-pass overwrite-then-unlink; every `secure_delete_tree` cap enforced before any overwrite; directory removal asserted by `stat`, not by `file_exists()` (which reports False for any directory)
- [ ] **Manual on-device validation (release gate).** Includes running a real `platform.wipe()` and verifying after reboot that the internal storage is actually gone — a failure there is invisible to anything that does not survive a power cycle. Every test here is host-side: `erase_and_format` runs against a fake block device and the tree tests run against the simulator's `fs/` directory. The part that needs real hardware — `erase_and_format()` against a real `pyb.SDCard`, actual block geometry, throughput over a full card, GUI responsiveness during the erase, and `os.VfsFat.mkfs()` afterwards — has not been exercised. #347 (hardware-in-the-loop) would be the eventual home for automating this.

## Notes

- Best-effort, not forensic sanitization — now stated in the code, not just here. SD cards do their own wear levelling: a logically overwritten block may still exist physically in controller-managed space. A software overwrite is the strongest erase possible in software, but against a well-equipped forensic attacker, physical destruction is the only certain way to sanitize any SD card.
- The full-card erase and `secure_delete_file()` both write zeros, for the RNG-throughput reason above. `platform.wipe()` is now the only remaining random-overwrite path — there is no performance argument there (~100 KB), so it is purely the open uniformity question and is deliberately left to that decision.


schnuartz-ai:
Cross-references for reviewers:

- **#377** — standalone \
eturn in finally\ fix (uses \os.remove\). Minimal, ~120 lines.
- **#378** — full BitBox02 import/export + SD card menu feature (~5300 lines). This PR (#379) is a strict subset of #378.

See this PR's description for the full merge-order matrix (Options A/B/C) and the \do not merge independently without rebasing\ warning.

Schnuartz:
@al-munazzim Would you think this is ready to merge?

al-munazzim:
Thanks for the ping. I would not merge this yet because the `native-tests` check is currently failing.

I reproduced the failure locally on the PR branch with `python test/run_native_tests.py`: the new `SecureDeleteTreeTest` cases fail before exercising `secure_delete_tree()` because `setUp()` only calls `platform.maybe_mkdir("testdir/deltree")`. `maybe_mkdir()` is a single-level `os.mkdir()`, so when `testdir` does not already exist the nested mkdir is swallowed and the first file creation raises:

```text
FileNotFoundError: [Errno 2] No such file or directory: 'testdir/deltree/file_0.bin'
```

A minimal fix that makes the full native suite pass locally is:

```diff
diff --git a/test/tests_native/test_platform_sd.py b/test/tests_native/test_platform_sd.py
@@
 class SecureDeleteTreeTest(TestCase):
     def setUp(self):
         clear_testdir()
+        platform.maybe_mkdir("testdir")
         platform.maybe_mkdir("testdir/deltree")
         self.path = "testdir/deltree"
 
     def tearDown(self):
-        platform.delete_recursively("testdir", include_self=True)
+        try:
+            platform.delete_recursively("testdir", include_self=True)
+        except OSError:
+            pass
```

After applying that locally I got:

```text
python -m compileall src test  # OK
Ran 18 tests in 0.348s
OK
```

I tried to push the fix directly to the PR branch, but GitHub denied write access to both `Schnuartz/specter-diy` and `cryptoadvance/specter-diy` for my key. If you apply the small test setup fix and CI goes green, I can take another look at merge readiness.

schnuartz-ai:
@al-munazzim Applied your suggested test-setup fix as-is in d1345ea. Full native suite passes locally (Ran 18 tests — OK, plus compileall clean), and CI is green on the new head: native-tests, tests, and Build Firmware all pass. Could you take another look at merge readiness?

Schnuartz:
@al-munazzim can you review again?

Schnuartz:
@al-munazzim please review again

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
     def __exit__(self, *args, **kwargs):
         self.unmount()
 
+    async def erase_and_format(self, progress_cb=None):
+        """
+        Securely erases the SD card - overwrites every block with random
+        data, the same approach platform.wipe() uses for the internal
```
maggo83: is reference to platform.wipe() needed here?
in https://github.com/cryptoadvance/specter-diy/pull/386 wipe will write 0s and not random data.
lets try to avoid drift by excessive cross-referencing by copy/paste/repeat

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
...
+        flash - and then creates a fresh, empty FAT filesystem on it.
+
+        This is irreversible and destroys EVERYTHING on the card, not
+        just files Specter-DIY created. Do not cancel or reset the device
```
maggo83: Part of this seems to be user (not developer) facing hints/comments. Should be put onto user facong messages not in the comment of code

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
     def __exit__(self, *args, **kwargs):
         self.unmount()
 
+    async def erase_and_format(self, progress_cb=None):
```
maggo83, line 258: I am not super familiar with this module yet, but I'd understand "self" by default is the whole specter platform in this module. But then equating self.is_present with the SD card being present does not make sense.
So I'd propose to change either "self" to a name that indicates we talk about the SD-Card or to change is_present to indicate it is relsting to the SD-Card only (e.g. as opposed to the SmartCard that can be present or not)

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
...
+        the whole operation. The event loop is allowed to run after every
+        chunk either way (asyncio.sleep_ms(0)) - also when a progress_cb
+        is given, since a callback that never awaits anything itself
+        (like a simple screen redraw) would otherwise starve the GUI's
+        update loop for the entire operation.
+        """
+        if not self.is_present:
+            raise RuntimeError("SD card is not present")
+        self.unmount()
+        if self._sd is None:
```
maggo83: Can't we use a clear marker that says "simulator"? inferring _sd=none means simulator is not obvious. And I think we already have somewhere one clear marker that says simulator or not. If not we should create it

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
...
+                self._sd.power(True)
+                block_size = self._sd.ioctl(5, None)
+                block_count = self._sd.ioctl(4, None)
+            except OSError as e:
+                raise RuntimeError(
+                    "Could not access the SD card before secure erase "
+                    "(card may have been removed):\n\n%s" % e
+                ) from e
+            if (
+                not isinstance(block_size, int)
```
maggo83: I think this kind of sanity check on blokc_size and _count is (or should) also be reused in other places, e.g. when clearing/deleting the internal flash and QSPI, see e.g. https://github.com/cryptoadvance/specter-diy/pull/386

-> let' not duplicate that everytime but make it a small helper function

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
...
+                        "The card is now in a half-overwritten, unusable "
+                        "state and must be reformatted before it can be "
+                        "used again." % e
+                    ) from e
+                # MicroPython block devices conventionally return None (or
+                # 0), while the STM32 binding used here returns True/False.
+                # Accept only those known success values; any other result
+                # must fail closed so a missed chunk can never be reported
+                # as a successful secure erase.
+                if not (
```
maggo83: since result is not initialized before the "try", it may either be undefined or still have the value of the last iteration of the for loop

src/platform.py
```diff
@@ -113,6 +143,133 @@ def __enter__(self):
...
+                try:
+                    if progress_cb is not None:
+                        await progress_cb((start + n) / block_count)
+                    # Always yield to the event loop, even with a
+                    # progress_cb: a callback that never awaits (e.g. one
+                    # that only redraws a progress bar) runs synchronously
+                    # and would otherwise keep every other task - including
+                    # the GUI update loop - from running until the whole
+                    # card is overwritten.
+                    await asyncio.sleep_ms(0)
```
maggo83, line 403: should this be in another try/except block to ensure GUI also gets called if progress_cb is faulty/fails?

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
     raise RuntimeError("Failed to delete folder %s" % path)
 
 
+def secure_delete_file(path, passes=3, max_size=None):
```
maggo83: Do we really need several passes? Seems to be security theater.

Simple internet search delivers::
A single overwrite pass is sufficient for securely wiping an SD card. 

According to NIST SP 800-88 guidelines, multiple passes are not required for modern flash storage media, including SD cards.  The primary justification for multi-pass wiping (such as the 35-pass Gutmann method) was based on older magnetic media where residual magnetic signals could potentially be recovered; this does not apply to NAND flash memory. 

Key considerations for SD card erasure include:

Wear and Tear: Performing multiple passes unnecessarily increases write cycles, which degrades the limited lifespan of the flash memory cells. 
Ineffectiveness: Additional passes do not significantly improve security or prevent data recovery on flash storage compared to a single pass. 
Method: Because SD cards typically lack TRIM support and do not accept ATA Secure Erase commands, a single full-drive overwrite using software is the recommended approach

Also for internal Flash/QSPI we deem one pass sufficient:
https://github.com/cryptoadvance/specter-diy/pull/386

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+    """
+    Overwrites a file's contents with fresh random data `passes` times,
+    syncing after each pass, before deleting it - the same
+    overwrite-then-unlink principle BitBox02's firmware uses when it
+    replaces a backup file (_delete_file() / sd_erase_file_in_subdir() in
+    bitbox02-firmware/src/sd.c), except that overwrites once with a fixed
+    byte (0xAC); this uses fresh random data on every one of several
```
maggo83: I think we should have a uniform approach to wether we overwrite with a constant or woth random data on case of wipes. I currently don't see why this should differ depending on the device, compare e.g. https://github.com/cryptoadvance/specter-diy/pull/386

src/platform.py
```diff
@@ -312,6 +396,103 @@ def delete_recursively(path, include_self=False):
...
+    stat-then-open per pass: a stat/open gap would let an attacker with
+    write access swap the path (e.g. via a hardlink on FAT variants that
+    support them) between the size check and the overwrite, causing us
+    to wipe the wrong file. The size is read from the same handle via
+    seek(0, 2)/tell(), and that handle is used for every overwrite pass.
+    """
+    with open(path, "r+b") as f:
+        f.seek(0, 2)  # seek to end
+        size = f.tell()
+        for _ in range(passes):
```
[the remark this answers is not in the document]
maggo83: Alternatively: hardcode to only do one pass as several passes might be security theater

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+    write access swap the path (e.g. via a hardlink on FAT variants that
+    support them) between the size check and the overwrite, causing us
+    to wipe the wrong file. The size is read from the same handle via
+    seek(0, 2)/tell(), and that handle is used for every overwrite pass.
+    """
+    _validate_secure_delete_passes(passes)
+    with open(path, "r+b") as f:
+        f.seek(0, 2)  # seek to end
+        size = f.tell()
+        if max_size is not None and size > max_size:
```
maggo83: how is max_size? chosen? is this really a run-time error or would a fitting return value be sufficient?

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+# for an unbounded amount of time (DoS). A BitBox backup directory has only a
+# few entries, so this leaves ample room for legitimate trees.
+SECURE_DELETE_MAX_FILES = 200
+SECURE_DELETE_MAX_ENTRIES = SECURE_DELETE_MAX_FILES
+SECURE_DELETE_MAX_DEPTH = 16
+SECURE_DELETE_MAX_FILE_BYTES = 4 * 1024 * 1024
+SECURE_DELETE_MAX_TOTAL_BYTES = 8 * 1024 * 1024
+
+
+def secure_delete_tree(path, passes=3):
```
maggo83: do we really need more than one pass?

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+    than SECURE_DELETE_MAX_FILE_BYTES, more than
+    SECURE_DELETE_MAX_TOTAL_BYTES of file data, or a depth greater than
+    SECURE_DELETE_MAX_DEPTH is rejected up front. This prevents a partial
+    wipe and stops an adversarial directory from forcing unbounded traversal,
+    recursion, or overwrite work. The caller can offer "Format entire SD
+    card" as the fallback for a tree that exceeds a cap.
+    """
+    _validate_secure_delete_passes(passes)
+    files = []
+    _collect_files(
```
maggo83: passing files [] as input and having it filled by the function and then use it as output might _technically_ work but is anything but obvious.
It is way more easy to understand (and maintain) to just have the file list be a returned output of collect_files.

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+        path,
+        files,
+        SECURE_DELETE_MAX_ENTRIES,
+        SECURE_DELETE_MAX_DEPTH,
+        max_file_bytes=SECURE_DELETE_MAX_FILE_BYTES,
+        max_total_bytes=SECURE_DELETE_MAX_TOTAL_BYTES,
+    )
+    remaining_bytes = SECURE_DELETE_MAX_TOTAL_BYTES
+    for full in files:
```
maggo83: should "full" not be rather called "file"? What does full signify here?

src/platform.py
maggo83: okay... this comes from secure_delete_tree. I think it would be better to have the error reaction then in secure_delete_tree. the meaning/significance of this error is unclear in secure_delete_file

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+        size = secure_delete_file(
+            full,
+            passes=passes,
+            max_size=min(SECURE_DELETE_MAX_FILE_BYTES, remaining_bytes),
+        )
+        remaining_bytes -= size
+    _remove_empty_dirs(path)
+
+
+def _collect_files(
```
maggo83: Recursion in constrained systems: not so good. Isn't an iterative approach better suited to our environment?

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+                if max_file_bytes is not None or max_total_bytes is not None:
+                    try:
+                        size = os.stat(full)[6]
+                    except Exception as e:
+                        raise RuntimeError(
+                            "Could not inspect file before secure delete: %s"
+                            % full
+                        ) from e
+                    if (
+                        not isinstance(size, int)
```
maggo83: let's use a common helper function for this type of sanity check

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+                    max_total_bytes=max_total_bytes,
+                )
+    finally:
+        close = getattr(entries, "close", None)
+        if close is not None:
+            close()
+    return entry_count, total_bytes
+
+
+def _remove_empty_dirs(path, max_depth=SECURE_DELETE_MAX_DEPTH, depth=0):
```
maggo83: again: iterative approach should be preferred over recursive on constrained devices

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+                    max_depth,
+                    depth=depth + 1,
+                    entry_count=entry_count,
+                    total_bytes=total_bytes,
+                    max_file_bytes=max_file_bytes,
+                    max_total_bytes=max_total_bytes,
+                )
+    finally:
+        close = getattr(entries, "close", None)
+        if close is not None:
```
maggo83: nit proposal:
if callable(close):

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+        )
+    entries = os.ilistdir(path)
+    try:
+        for name, entry_type, *_rest in entries:
+            if name in (".", ".."):
+                continue
+            if entry_type == 0x4000:
+                _remove_empty_dirs("%s/%s" % (path, name), max_depth, depth + 1)
+    finally:
+        close = getattr(entries, "close", None)
```
maggo83: nit proposal:
if callable(close):

src/platform.py
```diff
@@ -312,6 +469,228 @@ def delete_recursively(path, include_self=False):
...
+            "card' instead" % max_depth
+        )
+    entries = os.ilistdir(path)
+    try:
+        for name, entry_type, *_rest in entries:
+            if name in (".", ".."):
+                continue
+            if entry_type == 0x4000:
+                _remove_empty_dirs("%s/%s" % (path, name), max_depth, depth + 1)
+    finally:
```
maggo83: The logic to safely close all entries is duplicated between remove empty dirs and collect files already. consider putting it into a helper function

test/tests_native/test_platform_sd.py
maggo83, line 1: General question:
Test do not seem to care/log if run on a simulator or a real device/HW.
We should be able to tell from testlogs where it ran and also ensure before merge/release that these tests ran on real hw at least once.
How do we handle that?

src/platform.py
schnuartz-ai: Removed. The docstring no longer mentions `platform.wipe()` at all.

src/platform.py
schnuartz-ai: Agreed. The docstring is developer-facing again; the "do not cancel / do not remove the card" wording is now explicitly the caller's responsibility (the UI that drives the erase), and the actionable recovery text stays only in the RuntimeError messages, which do reach the user.

src/platform.py
schnuartz-ai: `is_present` is a property of `class SDCard`, so `self` here is the card, not the platform module - `platform.sdcard.is_present`. No rename needed, but the neighbouring `has_block_device` property added in this push should make the class context more obvious at the call site.

src/platform.py
schnuartz-ai: Right, the module-level `simulator` flag at platform.py:8 is that marker. Branching directly on it would break the native tests, which construct a real `SDCard` with a fake block device while running on the simulator build. So the condition is now a named property, `has_block_device`, whose docstring points at `simulator` and explains that the simulator build is the case where SDCard is constructed without a block device. No more inferring anything from `_sd is None` at the call site.

src/platform.py
schnuartz-ai: Done - extracted as `validate_block_geometry(block_size, block_count, device)`, backed by `is_valid_count()`. `SDCard.erase_and_format()` and `platform.wipe()` both use it now; `wipe()` previously validated nothing at all.

src/platform.py
schnuartz-ai: `result = None` before the try now. It was in fact unreachable-if-stale (the `except OSError` re-raises, so the check below never ran on a failed write), but the initialisation makes that obvious instead of something a reader has to prove.

src/platform.py
schnuartz-ai: Yes - split. `progress_cb` now has its own try: a `CancelledError` from it still reports an interrupted wipe, any other exception is captured and logged, and the `asyncio.sleep_ms(0)` yield that keeps the GUI task alive runs either way.

src/platform.py
schnuartz-ai: You are right, and I have taken it out. It is one pass now, and the docstring states why (NIST SP 800-88: multi-pass is a magnetic-media practice; on flash the extra passes only spend write cycles and cannot reach retired physical blocks on wear-levelled media anyway).

src/platform.py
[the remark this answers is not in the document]
schnuartz-ai: Done, and gone further than hardcoding: the `passes` parameter and its validator are removed entirely, so no caller can downgrade or inflate the primitive.

src/platform.py
schnuartz-ai: No - single pass now, see the reply above.

src/platform.py
schnuartz-ai: Agreed on uniformity, and this PR is already consistent with what is in master: `platform.wipe()` on master overwrites the internal flash with `os.urandom()`, and both the SD erase and the file overwrite here use random data too. #386, which proposed switching to zeros, was closed unmerged, so there is currently nothing pulling the other way. If you would rather standardise on a constant byte, that is a single change in both places - happy to do it, but it should be one decision applied to flash, QSPI and SD together rather than something this PR does unilaterally.

src/platform.py
schnuartz-ai: Both of your points taken: `max_size` is removed from `secure_delete_file()` entirely. All cap decisions - per-file, whole-tree and entry-count - now live in `secure_delete_tree()`/`_collect_files()`, where the tree is inspected up front and where the "use Format entire SD card instead" hint actually makes sense. `secure_delete_file()` just overwrites the file it is handed.

schnuartz-ai: Exactly - moved. See the reply on the other thread: `max_size` is gone from `secure_delete_file()` and the error is raised in `_collect_files()` before anything is overwritten.

src/platform.py
schnuartz-ai: Agreed - `_collect_files()` now builds and returns the list itself. No in/out parameter.

src/platform.py
schnuartz-ai: Renamed to `file_path`. "full" was short for "full path" and carried no other meaning.

src/platform.py
schnuartz-ai: Rewritten iteratively with an explicit `stack` of `(path, depth)` pairs. Same for `_remove_empty_dirs()`.

src/platform.py
schnuartz-ai: Done - also iterative now. It collects the directories deepest-last on the stack walk and then removes them in reverse, so children are always gone before their parent.

src/platform.py
schnuartz-ai: Done. `is_valid_count(value, allow_zero=...)` is the shared primitive: it rejects bools and non-integers once, and is used by `validate_block_geometry()` (SD erase + flash wipe) and by `_checked_file_size()`.

src/platform.py
schnuartz-ai: Applied.

src/platform.py
schnuartz-ai: Applied.

src/platform.py
schnuartz-ai: Factored into `_close_entries(entries)`, used by both traversals. It could not become a "return the whole listing" helper: the caps have to abort mid-enumeration, so an adversarial directory is never fully walked or retained - hence a close helper rather than a list helper.

test/tests_native/test_platform_sd.py
schnuartz-ai: Fair, and I do not think this PR can answer it on its own - it is a repo-wide gap. Where things stand concretely:

The tests in this file are pure host-side unit tests. `EraseAndFormatTest` drives a fake block device, so it never touches an SD card, and the `SecureDelete*` tests run against the simulator's `fs/` directory on the host filesystem. They are meaningful for the control flow (caps enforced before any overwrite, event-loop yields, error translation, no silent success), and meaningless as evidence about real hardware. Nothing here is capable of running on device, so a "where did this run" log line would only ever say "host".

What is genuinely untested is the part that needs a card: `erase_and_format()` against a real `pyb.SDCard` - block geometry from the actual ioctls, write throughput and timing over a full card, whether the GUI stays responsive for the whole erase, and `os.VfsFat.mkfs()` afterwards. #347 (hardware-in-the-loop testing) looks like the right home for that.

For this PR I would suggest treating a manual on-device run of the SD format and of a keystore-file delete as a release gate, the same way #386 was treated - I have flagged it in the PR description under Test plan. If you would rather I add an explicit `@skip` marker plus a "host-only" annotation to these test classes so the distinction is visible in the log, say the word and I will.

schnuartz-ai:
@maggo83 Thanks - all of your points are addressed in 938f96b, and I have replied inline on each thread. Summary of what changed:

**Security substance**

- **One overwrite pass, not three.** You were right that the extra passes are theatre on flash. The `passes` parameter and its validator are removed entirely rather than defaulted to 1, so a caller cannot downgrade or inflate the primitive. The docstring now carries the NIST SP 800-88 reasoning.
- **Random vs. constant:** left as random, because that is what `platform.wipe()` on master already does for the internal flash - so this PR is consistent with what is in the tree today. #386, which proposed zeros, was closed unmerged. If you want to standardise on a constant byte, it should be one decision applied to flash, QSPI and SD together; happy to make that change, just not unilaterally from this PR.

**Structure**

- `_collect_files()` and `_remove_empty_dirs()` are **iterative** now (explicit stack), no recursion on the constrained target.
- `_collect_files()` **returns** its list instead of filling a caller-supplied one; the per-entry variable is `file_path`, not `full`.
- `max_size` is gone from `secure_delete_file()`. Every cap decision - per file, whole tree, entry count, depth - now lives in `secure_delete_tree()`/`_collect_files()`, where the tree is actually inspected and where the "use Format entire SD card instead" hint means something.

**Shared helpers instead of copy/paste**

- `is_valid_count()` + `validate_block_geometry()` are used by the SD erase, by `_checked_file_size()`, and by `platform.wipe()` - which previously validated no geometry at all.
- `_close_entries()` replaces the duplicated ilistdir close logic, with `callable(close)` as you proposed.

**Robustness / readability**

- A failing `progress_cb` no longer skips the event-loop yield; it is caught and logged separately, while a `CancelledError` from it still reports an interrupted wipe.
- `result` is initialised before the `writeblocks()` try block.
- The simulator branch tests a named `SDCard.has_block_device` property that points at the module-level `simulator` flag, instead of `_sd is None` at the call site.
- Docstrings dropped the `platform.wipe()` and bitbox02-firmware cross-references, and the user-facing "do not cancel" wording - that belongs in the UI, not in a comment.

**Tests:** `SECURE_DELETE_MAX_FILES` renamed to `SECURE_DELETE_MAX_ENTRIES`; the pass-count and `max_size` tests are replaced by one asserting the single-pass overwrite-then-unlink behaviour. `Ran 34 tests ... OK` locally plus `compileall` clean.

The Copilot findings from the earlier round (geometry validation, ioctl error translation, cancellation reporting, short-write and strict-sync handling, cap enforced during enumeration rather than after) were already addressed in 25746f4 / 55b9bb2 / 32c43f8.

On your hardware-testing question - answered at length on that thread. Short version: nothing in this file can run on device, and the part that genuinely needs a card (`erase_and_format()` against a real `pyb.SDCard`) is untested. I would treat a manual on-device run as a release gate, as was done for #386, with #347 as the eventual home for automating it.

src/platform.py
maggo83: @miketlk : do you have any opinion on that?
if one pass is sufficient, an attacker cannot infer much from reading all 0s, while all 0s compared to random data is easier to implement (no dependency on rng still working) and verify (easy constant test pattern to compare to).

Is there any deeper reasoning to use random data instead of 0s? Maybe something related to steganography?

Schnuartz:
@al-munazzim can you review pls

schnuartz-ai:
Follow-up review pass found three things worth fixing. All addressed in 273847f.

### 1. `unmount()` cut power while `/sd` was still mounted (the real one)

When `os.umount("/sd")` failed, the old code kept `_mounted = True` so a later cleanup could retry — but powered the block device off anyway. That is an incoherent state: VFS still has `/sd` mounted, Specter still thinks the card is mounted, and the interface underneath is dead. Worse, the retry path starts with `os.sync()`, so the retry would have synced against hardware that could no longer answer.

Power is now cut only once the VFS mount is actually gone, or once the card is no longer there to retry against:

```python
if unmounted or not self._present_or_gone():
    self._sd.power(False)
```

`_present_or_gone()` wraps the presence probe so a probe that itself raises cannot mask the umount error already being reported, and treats a card we cannot interrogate as gone (holding an unresponsive interface powered would serve no purpose).

Four tests cover it, including the full retry path — failure keeps the card powered, a second `unmount()` then succeeds and powers down. Both new invariants fail against the previous unconditional power-off; I verified that by reverting the guard and re-running.

### 2. `os.VfsFat.mkfs()` was effectively untested

The native stub swallowed the call, so nothing checked it. Added three tests via a recording stub: `mkfs()` is called exactly once, against the same block device, and only after every block was written; it is **not** called when the overwrite fails partway (a half-overwritten card must never be handed a fresh filesystem that presents it as clean); and an `OSError` from it produces the "overwrite completed, no valid filesystem" `RuntimeError` with the card powered down.

### 3. Two tree-removal assertions could not fail

`test_normal_tree_deleted` and `test_tree_at_cap_deleted` asserted `assertFalse(platform.file_exists(self.path))` — but `file_exists()` works by opening the path as a regular file, so it returns `False` for *any* directory, present or not. Those assertions were vacuous.

They now `os.stat()` and check the parent listing, plus a nested-tree case. Confirmed as real regression tests: stub out `_remove_empty_dirs()` and all three fail; under the old assertion they passed.

No deletion bug was hiding behind them — `_remove_empty_dirs()` does walk iteratively and `os.rmdir()` in reverse order correctly. The tests just could not have caught one.

### Also

Corrected the `writeblocks()` comment. The STM32 binding returns the HAL status as an integer (0 = success), not `True`/`False`. The accepted-value check already handled that — only the comment was wrong.

**Tests:** `Ran 41 tests ... OK` (was 34), `compileall` clean.

### Still open, and not addressed here

- **Hardware validation remains the release gate.** Everything above is still host-side. `erase_and_format()` against a real `pyb.SDCard` — actual ioctl geometry, repeated 128 KiB allocations on the real MicroPython heap, full-card write reliability and duration, GUI responsiveness across a multi-minute erase, physical card removal mid-erase, the real `mkfs()`, and whether the freshly formatted card mounts and round-trips afterwards — has not been demonstrated. This is called out unchecked in the PR description and I would not merge without it.
- **Random vs zeros** is still open — @maggo83 asked @miketlk for an opinion. Replied separately on that thread.

src/platform.py
schnuartz-ai: Adding one correction to what I wrote earlier, since this thread is now waiting on @miketlk.

I cited NIST SP 800-88 Rev. 1. The current document is **Rev. 2, published September 2025**, which supersedes it. It does not change the single-pass conclusion, and it is explicit on two points that matter here:

- overwrite-based clearing means replacing user-addressable data with **non-sensitive** data — it does not require that data to be random;
- repeated passes buy little on flash, and wear levelling means a logical overwrite cannot reach every historic physical copy regardless.

So by the standard, one pass of zeros and one pass of random are equally acceptable. This is a preference question, not a security one.

If it helps break the tie, I would lean **zeros**, for reasons that are all practical rather than cryptographic: deterministic and trivially verifiable on-device or with a card reader, no dependence on RNG availability at wipe time, and no entropy-generation work for zero security benefit on a device where the TRNG matters elsewhere.

I left it as random here only because `platform.wipe()` on master already uses `os.urandom()` for the internal flash, and I did not want this PR to unilaterally create the very inconsistency you were asking about. Whichever way you and @miketlk decide, I am happy to implement it — but it should land as one change across internal flash, QSPI and SD rather than SD alone. Say the word and I will do it in this PR or a follow-up, whichever you prefer.

(Note #386, which proposed zeros for the QSPI wipe, was closed unmerged on 2026-08-28, so there is currently nothing else in flight pulling toward zeros.)

schnuartz-ai:
Good catch on `power()` — I followed the pins and confirmed it against the actual source rather than taking it on trust. Fixed in 28382d4.

### Verified against the pinned firmware

`specter-diy` → `f469-disco` **db3ce3e** → `micropython` **6bdf1b6**, `ports/stm32/sdcard.c`:

```c
STATIC mp_obj_t sd_power(mp_obj_t self, mp_obj_t state) {
    bool result;
    if (mp_obj_is_true(state)) {
        result = sdcard_power_on();
    } else {
        sdcard_power_off();
        result = true;
    }
    return mp_obj_new_bool(result);
}
```

So a failed power-on returns `False` and raises nothing — exactly as described. And the downstream path is confirmed too: `sdcard_get_capacity_in_bytes()` returns `0` for anything other than `PYB_SDMMC_FLAG_ACTIVE | ..._SD`, so `ioctl(4)` reports 0 blocks and the failure surfaced as `invalid geometry`.

Fail-safe — nothing was ever overwritten — but the wrong diagnosis for a destructive operation. Now:

```python
powered = self._sd.power(True)
if powered is False:
    raise RuntimeError(
        "Could not access the SD card before secure erase "
        "(the card could not be initialized)."
    )
```

Only an explicit `False` counts; block devices that report no status return `None`, and there is a test pinning that so the guard cannot drift into rejecting them.

Two tests: power-on failure produces the initialization error rather than the geometry one, with zero `writeblocks()` and zero `mkfs()` calls and the card still powered down on the way out; and a `None` return still completes a normal erase. The first fails if the guard is removed — verified.

### The `writeblocks()` comment: I had it wrong, in the other direction

You are right, and my previous commit made this worse rather than better. The pinned revision really does return a bool:

```c
mp_uint_t ret = sdcard_write_blocks(...);
return mp_obj_new_bool(ret == 0);
```

The integer-HAL-status wording I introduced describes generic MicroPython, not this pin. The original True/False comment was correct and I "corrected" it into being wrong. The comment now names both cases and says which one is this firmware. The accepted-value check itself was always right — it accepts `None`, `True` and integer `0` — so no functional change, but the comment misled twice and that is worth not repeating.

### Not changed: `mount()`

`SDCard.mount()` has the same pattern — `self._sd.power(True)` with the return value discarded, then `os.mount()`. It is pre-existing master code that this PR does not otherwise touch, so I have left it rather than widen the diff. There a failed power-on surfaces as an `os.mount()` error, which is at least not actively misleading. Happy to fix it here or in a follow-up, whichever you prefer.

**Tests:** `Ran 43 tests ... OK` (was 41), `compileall` clean.

### Still the only thing left

Hardware validation. Unchanged and still unchecked in the description. Worth noting that this finding is a good argument for it: the failure mode was invisible to the host suite precisely because the fake block device modelled power failure as an exception, which is not what the real driver does. Same category of gap applies to the repeated 128 KiB `os.urandom()` allocations on a fragmented MicroPython heap — no desktop fake can tell us anything about that.

For the on-device session I would run the sequence you listed: known files → full erase → UI responsive throughout → progress reaches 100% → card FAT-formatted → remount on Specter → write, read back → reboot and confirm still usable → separately, SD mnemonic-file deletion → and if practical, pull the card mid-wipe once to exercise the interrupted path and the `unmount()` retry behaviour changed in 273847f.

schnuartz-ai:
All four points are valid. Fixed in 6db0599.

### 1. `MemoryError` in the wipe loop — agreed, this was a merge fix

Confirmed the mechanism at the pinned revision: `os_urandom()` in `ports/stm32/moduos.c` builds a fresh `vstr` per call, so the old code really did request a new 128 KiB object on every chunk. `MemoryError` is not an `OSError`, so it escaped the handler as a raw traceback — after blocks were already overwritten, with the card powered off in `finally` and no "half-overwritten, reformat it" message. Exactly the failure class the rest of this code guards against.

The buffer is now allocated **once, before the first destructive write**, and reused:

```python
try:
    buf = bytearray(chunk_blocks * block_size)
except MemoryError as e:
    raise RuntimeError(
        "Not enough memory to start the secure erase. No data "
        "has been changed - reboot the device and try again."
    ) from e
```

Failing there costs nothing, because nothing has been touched yet. Per chunk it is refilled in place by `fill_random()` from small scratch reads, and the short final chunk uses a `memoryview` slice rather than a copy. The loop also translates `MemoryError` into the interrupted-wipe error, since the scratch reads can still fail on an exhausted heap.

I kept random rather than pre-empting the zeros discussion — with a reusable buffer the allocation argument no longer favours one over the other, so that stays purely maintainers' choice. If you go with zeros it collapses to a single `bytearray()` and `fill_random()` disappears.

A test asserts no allocation inside the loop exceeds the scratch size, so this cannot silently regress to per-chunk allocation.

### 2. Geometry validation was incomplete — agreed

`wipe()` writes a hardcoded range, so a device reporting `block_count = 300` passed the check and was then written up to block 449, past its own reported geometry. Now:

```python
validate_block_geometry(block_size, block_count, "internal flash",
                        min_blocks=WIPE_LAST_BLOCK + 1)
```

and the range is named (`WIPE_FIRST_BLOCK` / `WIPE_LAST_BLOCK`) and used by the loop, so the range written and the minimum required to write it cannot drift apart. Not exploitable on the current target as you say — but the abstraction now actually holds.

### 3. Cleanup masking the real error — agreed, and it was inconsistent

You are right that this contradicted the `unmount()` work sitting right next to it. `erase_and_format()`'s `finally` now follows the same rule: a cleanup error is surfaced only when nothing more important is in flight, otherwise logged. Two tests — one where the write error must survive a failing power-off, one where a cleanup failure on the success path must not be swallowed either.

### 4. "128 KiB bounded" was only true for ordinary block sizes — agreed

`chunk_blocks` falls to 1 for a large `block_size`, so the buffer was one block and the comment was aspirational. `MAX_SANE_BLOCK_SIZE` (64 KiB) is now part of `validate_block_geometry()`, which makes the bound real and the comment true. Real cards report 512.

**Tests:** `Ran 55 tests ... OK` (was 43). Each new test verified to fail against the previous behaviour.

### Still the only blocker

Hardware validation, unchanged. Point 1 is a good illustration of why: the heap-fragmentation failure it describes is precisely what a desktop fake block device cannot reproduce, and the fix for it — a single up-front allocation — is itself only *plausible* until it has run on a real fragmented MicroPython heap. The on-device session should watch memory behaviour across a full-card erase specifically, not just the end result.

schnuartz-ai:
This is the right call and it settles the zeros-vs-random question for the full-card path on grounds that are not a preference. Fixed in 97598e8.

### Verified at the pinned revision

`ports/stm32/moduos.c` — one `rng_get()` **per byte**:

```c
STATIC mp_obj_t os_urandom(mp_obj_t num) {
    ...
    for (int i = 0; i < n; i++) {
        vstr.buf[i] = rng_get();
    }
```

and `ports/stm32/rng.c` — each of those busy-waits:

```c
// Wait for a new random number to be ready, takes on the order of 10us
while (!(RNG->SR & RNG_SR_DRDY)) { ... }
```

So ~3 h per GiB of pure RNG wait before a single SD write. Even taking the F4's best-case DRDY rate rather than the comment's 10 µs, a large card still lands in the hours. Your numbers hold, and the 256-byte scratch loop I added for heap safety made it worse in a second dimension: 512 `os.urandom()` calls per 128 KiB chunk, ~134 million across a 32 GiB card. I introduced that in 6db0599 while fixing the allocation problem and did not cost it out — that was my error.

### Fixed

The buffer is allocated once, stays zero-filled, and is never modified. `fill_random()` is gone entirely, along with the scratch loop and its failure mode. The erase loop is now:

```python
data = buf if n == chunk_blocks else view[:n * block_size]
result = self._sd.writeblocks(start, data)
```

The `MemoryError` guard stays: nothing in the loop allocates any more, but a driver may, and it still has to surface as an interrupted wipe rather than a raw traceback.

Two tests: a full erase makes **zero** `os.urandom()` calls and every written byte is zero; and the buffer is allocated exactly once at 128 KiB. The first fails if the loop ever goes back to random.

### Scope: full-card only, deliberately

Per your split, `secure_delete_file()` still overwrites with random and `platform.wipe()` is unchanged. At mnemonic-file sizes the per-byte RNG cost is irrelevant.

One number worth putting on the record for @maggo83 / @miketlk, since it bears on the uniformity decision: `secure_delete_tree()` caps a tree at `SECURE_DELETE_MAX_TOTAL_BYTES` = 8 MiB, and 8 MiB at ~10 µs/byte is on the order of **80 seconds** of RNG wait. Bounded and never reached by a real backup directory (a few KB), but it is the same root cause, and it is another argument for standardising on zeros everywhere rather than only here. I have not changed it, because that is the open question on that thread and it should be one decision across SD, internal flash and QSPI — not something this PR settles by side effect.

**Tests:** `Ran 54 tests ... OK`. (One fewer than before: the `fill_random` tests went with the function.)

### Remaining

Hardware validation, unchanged — and this finding sharpens what it needs to measure. A wall-clock time for a full erase is now a first-class acceptance criterion, not a nice-to-have: the whole reason this bug survived 55 green tests is that no host test has any notion of how long a write takes.

schnuartz-ai:
Fixed in 35cc12f — `Keep the temporary wipe buffer bounded.`

I grepped the rest of the module for the same staleness. The other "random" mentions are all still accurate:

- the block explaining *why* the erase does not use random (the RNG cost), which is the point of that comment;
- `secure_delete_file()`, which genuinely still overwrites with random data;
- `platform.wipe()`, likewise unchanged.

So that was the only leftover. `Ran 54 tests ... OK`.

schnuartz-ai:
This one is serious and you are right that it is a blocker. Fixed in 4890d16, and the branch is now updated onto current `master` (c615a74).

### Verified at the pinned revision — and it is worse than 1 second

`ports/stm32/flashbdev.c`:

```c
// If not a forced write, wait at least 5 seconds after last write to flush
if ((flash_flags & FLASH_FLAG_FORCE_WRITE) || HAL_GetTick() - flash_tick_counter_last_write >= 5000) {
```

Five seconds, not one. `wipe()` writes ~194 blocks and resets immediately, so it is not a race that might occasionally lose data — it is inside the window essentially by construction.

The rest of the chain confirms too:

- `pyb_flash_ioctl` case `MP_BLOCKDEV_IOCTL_SYNC` → `storage_flush()` (`ports/stm32/storage.c`)
- `storage_flush()` issues `BDEV_IOCTL_SYNC` **and** `BDEV2_IOCTL_SYNC`, so one call covers the QSPI device whose first blocks are inside the wiped range
- `machine_reset()` → `powerctrl_mcu_reset()` → `NVIC_SystemReset()`, with no `storage_flush()` anywhere on that path

So the device could come back up with the data still present, having told the user it was wiped. On the `CriticalErrorWipeImmediately` path that is the worst possible outcome.

And your second point is confirmed as well — `pyb_flash_writeblocks` returns `MP_OBJ_NEW_SMALL_INT(ret)`, an integer status, so every failed write was silent. The PR checked this carefully for `SDCard` and not at all here.

### Fixed

```python
result = f.ioctl(3, None)  # MP_BLOCKDEV_IOCTL_SYNC
if not is_block_op_success(result):
    raise RuntimeError(...)
```

after the last write, plus a fail-closed status check on every write. **A failed write or flush raises instead of rebooting** — the data is still there either way, and a loud failure beats a reset that presents the device as wiped. Worth flagging explicitly: callers in `specter.py` do not currently handle an exception from `wipe()`, so this surfaces as an unhandled error rather than a UI message. I have left that alone as a separate UI concern rather than widening this PR, but it should not stay that way.

The accepted-result logic is now one shared helper, `is_block_op_success()`, used by both destructive paths. The drivers genuinely disagree — `SDCard` returns `True`/`False`, `Flash` returns an integer, generic devices `None` — and that knowledge belongs in exactly one place.

### On scope

You framed it as revert-or-harden, and I went with harden. Reverting the geometry check would leave `wipe()` no worse than `master`, but it would also leave a destructive path this branch has already touched in a half-hardened state, which seemed like the worse of the two. Say the word if you would rather see `wipe()` split into its own PR — the change is self-contained and I can lift it out.

**Tests:** `Ran 61 tests ... OK`. Three new ones covering the ordering directly: sync happens after the last write and before reboot, the whole documented block range is written, and a failed write or failed sync aborts without rebooting. All three fail against the previous behaviour.

### Branch updated

Merged `upstream/master` (the #345 docs commit) rather than rebasing, so no published history was rewritten and no force-push was needed. CI is running on the merged head.

### Remaining

Hardware validation, unchanged — and this finding adds a specific item to it that was not on the list before: **run a real `platform.wipe()` and verify after reboot that the internal storage is actually gone.** That is the one case where "the tests pass" was furthest from "it works", since the failure mode is invisible to anything that does not survive a power cycle.

schnuartz-ai:
Items 2, 3 and 4 are fixed in 8de07f3. Item 1 stands and I cannot close it from here.

### 2 — `wipe()`'s failure contract now reaches the UI

You are right that leaving this as a follow-up was wrong for a security path, and you correctly quoted me saying it should not stay that way. `Specter.wipe_or_report()` catches the failure and says plainly:

> THE DEVICE WAS NOT WIPED. … Your data may still be present on this device. Do not treat it as erased, and do not pass it on or dispose of it in this state.

On the `CriticalErrorWipeImmediately` path a failure now **keeps offering the wipe** rather than continuing — the device still holds exactly what the wipe was meant to destroy, so falling through to normal operation is not an acceptable outcome. That also fixes a latent second bug: `handle_exception()`'s result is used as `await next_fn()`, and the old code only got away with returning nothing because `wipe()` never returned. Now that it can, returning would have crashed on `await None()`.

### 3 — the presence probe was backwards, and that was my error

`_present_or_gone()` treated "the probe raised" as proof the card was gone. That is not a proof of anything, and on the double fault you describe it cut power with `/sd` still mounted — recreating the exact state the surrounding fix exists to prevent. Renamed to `_card_is_absent()` and inverted: only a positive answer counts, unknown means still there. Worst case is a powered interface until a retry succeeds, which is the cheaper failure. The test asserting the old behaviour is inverted accordingly and fails against it.

### 4 — `secure_delete_file()` writes zeros

Fixed now rather than deferred, since #378 builds on this primitive. It is synchronous and yields to nothing, so at the 8 MiB cap the RNG cost would have blocked the firmware for over a minute — which undermines the caps, whose whole purpose is bounding that. Zeros match `erase_and_format()`, and NIST SP 800-88 asks for non-sensitive replacement data, not random data.

`platform.wipe()` is now the only remaining random-overwrite path. There is no performance argument there — ~100 KB, about a second — so it is purely the uniformity question on @maggo83's open thread, and I have deliberately not settled it here by side effect. **It is a two-line change whenever you and @miketlk decide.** My recommendation, for what it is worth: unify on zeros, since everything else already is.

### Test coverage — one honest gap

`Ran 62 tests ... OK`. The presence-probe and zero-overwrite changes both have tests that fail against the previous behaviour.

**`Specter.wipe_or_report()` has no automated coverage.** `specter.py` cannot be imported in the native harness (`microur` is absent), and stubbing enough GUI and host surface to import it would mean testing my stubs rather than the behaviour. Given that most of the real bugs this PR has surfaced — `power()` returning `False`, the RNG cost, the flash cache — survived precisely because a fake encoded my assumptions instead of the driver's behaviour, adding more fakes here seemed like the wrong lesson. It is verified by reading only, and it belongs on the manual pass.

### Remaining

Item 1, unchanged. The list has grown by one item that only hardware can answer: a real `platform.wipe()` followed by a reboot, proving the internal storage is actually gone — plus, now, confirming the failure screen renders correctly if a wipe does fail.

schnuartz-ai:
Finding 1 is a real bug and it is mine — I introduced the retry loop in 8de07f3 and never tested a second attempt. Fixed in f874350. Findings 2 and 3 are factually right; details below, including why the proposed read-back fix would not actually work on this driver.

### 1 🔴 — the retry could never succeed

Confirmed at the pin, `extmod/vfs.c`:

```c
if (vfs == NULL) {
    mp_raise_OSError(MP_EINVAL);
}
```

So the chain is exactly as you describe: attempt 1 unmounts both, the write fails, `wipe()` raises without rebooting and leaves them unmounted; attempt 2 dies on the first `os.umount()` and never reaches the overwrite. On the critical path that is an endless loop repeating "the device was NOT wiped" — the precise state the retry was added to escape.

`_umount_if_mounted()` now tolerates that one case and nothing else; any other unmount failure still propagates, because it means the filesystem is still attached and the wipe must not run.

**Your diagnosis of why the tests missed it is the important part.** `os.umount = lambda path: None` made unmount artificially idempotent — a fake encoding my assumption rather than the driver's behaviour. That is the third time in this review cycle the same failure mode has produced a real bug (`power()` returning `False`, the per-byte RNG cost, now this). The stub now raises `EINVAL` for an already-unmounted path like the real one, and there are tests for a successful retry after a failed write and for a genuine unmount error still aborting. The retry test fails with an error against the old code.

### 2 🟠 — you are right, and the proposed fix would not work either

Confirmed: `pyb_flash_ioctl` returns a hardcoded `MP_OBJ_NEW_SMALL_INT(0)` for SYNC, and `storage_flush()` discards the results of both underlying flushes. The check cannot detect a low-level failure on this firmware. I have stopped implying otherwise: issuing the sync is the part that matters and stays; the status check is kept as defensive handling for drivers that *do* report, and is now documented as exactly that.

On your read-back proposal — I liked it and went to implement it, then checked the driver:

```c
static uint8_t *flash_cache_get_addr_for_read(uint32_t flash_addr) {
    ...
    if (flash_cache_sector_id == flash_sector_id) {
        // in cache, copy from there
```

`BDEV_IOCTL_SYNC` clears `flash_flags` but **not** `flash_cache_sector_id`. So a read-back after the sync still reads from the RAM cache for the cached sector — and that is the most recently written one, i.e. exactly the sector most at risk. It would return the expected pattern whether or not anything reached the flash. That is worse than no verification, because it would look like proof. Noted in the code so nobody re-derives the idea and implements it.

Propagating the error properly means patching `pyb_flash_ioctl` / `storage_flush` in the MicroPython fork, which is a submodule change and well outside this PR.

### 3 🟠 — agreed, and now stated in the code

The caveat was only in the PR description. Both overwrite docstrings now say plainly that this is best-effort and not forensic sanitization, and why — wear levelling means an older physical copy can survive in space the controller no longer exposes, and NIST SP 800-88 says reaching every historic copy needs purge or destroy rather than an overwrite through the normal interface. No UI text in this PR claims otherwise (the SD menu lives in #378, which is worth checking on that point when it lands).

Crypto-erasure for seed material on SD is the right long-term answer and I agree it would remove the need to trust the controller at all. Out of scope here.

### 4, 5

Nothing to do — thank you for checking the BDEV1/BDEV2 boundary and the single-block write behaviour independently. That was a real risk and I had not verified it that way.

**Tests:** `Ran 64 tests ... OK` (was 62). CI running.

### Remaining

Item 1 of your previous list, unchanged: the hardware gate. This round is another argument for it — a bug that made the security retry path loop forever was invisible to 62 passing tests.

schnuartz-ai:
Pushed three commits (`eb96b4c`, `8552c9d`, `e6b2c03`) addressing a second round of review. Summary of what changed and, just as important, what I deliberately did **not** change.

### 1. A save that replaces an existing key file left the old one recoverable (`eb96b4c`)

This is the one that matters most, because it defeats the point of the branch in the workflow it is meant to protect.

`save_aead()` opens the path `"wb"`. On FAT that truncates the file, which frees its cluster chain **without overwriting it**, and the new contents are written wherever the allocator puts them. So:

```
save "reckless.mykey"   -> clusters 341-343   (ciphertext v1)
save over it            -> 341-343 freed, untouched; v2 written to 811-813
secure_delete_file()    -> overwrites 811-813 only
```

The v1 ciphertext survives in free space, and `enc_secret` — which decrypts it — is not rotated by a delete. Both `save_mnemonic()` implementations now `secure_delete_file()` the file they are about to replace, after the user confirms the overwrite, and abort the save if that overwrite fails rather than truncating over it anyway.

Writing the test for it surfaced a **pre-existing bug in master**: `FlashKeyStore.save_mnemonic()` references `Prompt` without importing it, so confirming an overwrite on internal flash raises `NameError` instead of asking. Fixed in the same commit (`SDKeyStore` imports it, which is why this never showed up there).

### 2. `mkfs()` after a full-card erase (`8552c9d`)

The 128 KiB erase buffer was still held while `os.VfsFat.mkfs()` allocated its own VFS/FatFs structures — the allocation most likely to fail on a fragmented heap, at the moment where failing costs most. Buffer released and `gc.collect()` before the format. `mkfs()` was also only guarded against `OSError`; a `MemoryError` there would have escaped as a raw traceback instead of "the data is gone, the filesystem is not".

### 3. `wipe()` writes zeros, and the guarantees are scoped (`e6b2c03`)

`wipe()` used `os.urandom()` per block. This is the emergency path — reached from `CriticalErrorWipeImmediately`, i.e. after something already went badly wrong — and it does not need the RNG peripheral, its driver and a per-block allocation standing between that state and the overwrite. NIST SP 800-88 asks for user-addressable data to be replaced with non-sensitive data, not for that data to be random, so the constant pattern is not weaker, and it matches `secure_delete_file()` / `erase_and_format()`. One zero block, allocated before the first write; a `MemoryError` there reports "the device has NOT been wiped". This closes the "zeros vs random" question the PR description left open — happy to revert that hunk if you prefer random here.

The docstrings now say what each operation does **not** cover:

* `wipe()` overwrites the internal flash in full, but only blocks 448-449 of the QSPI. The QSPI filesystem dies with them; the ~16 MiB behind it is **not** raw-overwritten and is protected by destroying the keys in internal flash — cryptographic erasure, not an overwrite. A hard reset also does not scrub SRAM.
* `secure_delete_file()` reaches the file's current allocation only: not earlier versions freed by a truncating rewrite (see 1), and not the directory entry, which FatFs marks rather than clears — so the file *name*, long-file-name records included, outlives its contents.
* `_strict_file_sync()` raises on every error the runtime reports, but on the pinned MicroPython the VFS discards the block device's write status (`// TODO handle error return` in `vfs_blockdev.c`) and `pyb.SDCard`'s `IOCTL_SYNC` is a no-op that always succeeds. A clean return is therefore not proof the overwrite reached the medium.
* `secure_delete_tree()` rejecting an over-cap tree before touching anything is a statement about the caps, not atomicity — an I/O error mid-loop still leaves part of the tree deleted.

### Deliberately not fixed here

* **Block-device error propagation in the MicroPython fork.** The gap under `_strict_file_sync()` is real: `pyb.SDCard.writeblocks()` correctly returns `False` on a rejected write, and the pinned VFS throws that away. Upstream MicroPython fixed this (`0` / negative errno contract), but backporting needs `vfs_blockdev.c` **and** the STM32 SD driver changed together — the old driver returns boolean `True`, which the new VFS logic would read as an invalid positive status. That belongs in the firmware fork, not in this PR.
* **Default filename = first BIP39 word.** `get_input(suggestion=self.mnemonic.split()[0])` puts a real seed word in the FAT directory entry, which (see above) survives a secure delete. Changing the default is a UX decision and pre-dates this branch, so I left it — but I think a neutral default is worth a follow-up.
* **SRAM scrub before `hard_reset()`.** Documented, not implemented.

### Tests

`Ran 68 tests ... OK` (host-side). New regression coverage: overwrite-before-replace for both keystores including the failed-overwrite abort, `Prompt` resolving in `FlashKeyStore`, `mkfs()` running out of memory reported as a completed erase, and `wipe()` writing zeros to every block in the documented range with zero `os.urandom()` calls.

The on-device release gate is unchanged and still open — none of this is validated on real F469 hardware yet, and for `wipe()` specifically that means a real power-cycle check.

src/keystore/flash.py
```diff
@@ -213,6 +214,27 @@ def fileprefix(self, path):
         return "specterdiy%s" % hexid
 
 
+    def _replace_existing(self, fullpath):
```
maggo83: This seems to be a simple wrapper around another function which a) does not add benefit/functionality and b) has unfitting name (replace implies delete and add new. here just the old is deleted).

src/keystore/sdcard.py
```diff
@@ -82,6 +82,18 @@ async def save_mnemonic(self):
                 if fullpath.startswith(self.sdpath):
                     platform.sdcard.unmount()
                 return
+            # See FlashKeyStore._replace_existing(): the old encrypted
+            # mnemonic has to be overwritten before the new file truncates
+            # it away, or it survives in free space.
+            try:
+                self._replace_existing(fullpath)
+            except Exception:
+                if fullpath.startswith(self.sdpath):
```
maggo83: why do we fall back to simple unmount w/o an error message if we think overwrite is needed for security?

src/platform.py
```diff
@@ -113,6 +245,219 @@ def __enter__(self):
...
 
+    @property
+    def has_block_device(self):
+        """
+        True when a real block device backs this card, so raw block
+        operations are possible. False in the simulator build (see the
+        module-level `simulator` flag), where SDCard is constructed
+        without a block device and /sd is an ordinary host directory.
+        """
+        return self._sd is not None
```
maggo83, line 256: The comment still suggest thisis used only to determine if we are running on a simulator. Not sure why we need a new function / logic for this, of we can just re-use the existing simulator flag

src/platform.py
```diff
@@ -113,6 +245,219 @@ def __enter__(self):
...
+            # 800-88 asks for user-addressable data to be replaced with
+            # non-sensitive data, not for that data to be random - and on
+            # this hardware random is not a viable option at card scale:
+            # os.urandom() calls rng_get() once per byte (moduos.c), and
+            # rng_get() busy-waits for RNG_SR_DRDY, "on the order of 10us"
+            # by its own comment. That is hours per gigabyte of pure RNG
+            # wait before a single SD write, so a full-card wipe would
+            # never finish in a usable time. Whole-card erase therefore
+            # writes zeros; see secure_delete_file() for the small-file
+            # path, where the same cost is irrelevant.
```
maggo83, line 351: @miketlk / @k9ert : we should have a general decision whether we overwrite with 0s or random data. Can you please chime in?

src/platform.py
```diff
@@ -113,6 +245,219 @@ def __enter__(self):
...
+            # Keep the temporary wipe buffer bounded. A 1 MB
+            # allocation is needlessly risky on a fragmented MicroPython
+            # heap; 128 KiB is still large enough to amortize SD writes
+            # while keeping the wipe usable on the target hardware. The
+            # geometry check above caps block_size, so this really is
+            # bounded even for a device reporting unusually large blocks.
+            chunk_blocks = max(1, (128 * 1024) // block_size)
+            # Allocate the write buffer ONCE, before the first destructive
+            # write, and reuse it for every chunk. Allocating per chunk
+            # meant a fragmented heap could fail the allocation partway
```
maggo83, line 333: Much duplication between this and following comments -> risk of drift. make sure the info is mentioned only in one place and in the one most suitbale. e.g. the comment regarding exception is best kept in the exception handling parts, not ahead here.

src/platform.py
```diff
@@ -312,6 +657,282 @@ def delete_recursively(path, include_self=False):
...
+      rewrite is then usually placed elsewhere. Overwriting the file that
+      exists today does not touch clusters an earlier version was written
+      to. A caller replacing a sensitive file should secure-delete it first
+      and create it fresh, rather than truncating over it.
+    * The directory entry. FatFs marks a deleted entry rather than clearing
+      it, so the file NAME - including its long-file-name records - stays
+      readable in the directory sector after the contents are gone.
+
+    Persistence of the overwrite is only as strong as _strict_file_sync()
+    below can make it on this runtime; see there.
```
maggo83, line 731: These comments seem like rationales that should rather go into a commit message or other documentation.

src/platform.py
```diff
@@ -312,6 +657,282 @@ def delete_recursively(path, include_self=False):
     raise RuntimeError("Failed to delete folder %s" % path)
 
 
+SECURE_DELETE_CHUNK = 4096
```
maggo83, line 679: magic number w/o rationale

src/platform.py
```diff
@@ -312,6 +657,282 @@ def delete_recursively(path, include_self=False):
...
+
+
+# An adversarial directory with thousands of entries would otherwise make
+# secure_delete_tree() enumerate, walk and remove entries for an unbounded
+# amount of time (DoS). A BitBox backup directory has only a few entries, so
+# these leave ample room for legitimate trees.
+SECURE_DELETE_MAX_ENTRIES = 200
+SECURE_DELETE_MAX_DEPTH = 16
+SECURE_DELETE_MAX_FILE_BYTES = 4 * 1024 * 1024
+SECURE_DELETE_MAX_TOTAL_BYTES = 8 * 1024 * 1024
```
maggo83, line 813: Are these values self determined? how? based on our hardware? direct copy/carry over from bitbox? what is the rationale?

src/platform.py
```diff
@@ -357,32 +978,156 @@ def reboot():
         pyb.hard_reset()
 
 
+# errno.EINVAL - same value in MicroPython's uerrno. Spelled out rather than
+# imported so this cannot depend on which errno module a build freezes in.
+_EINVAL = 22
```
maggo83, line 1022: why not simply import errno/uerrno? Grug not like magic number

src/platform.py
```diff
@@ -357,32 +978,156 @@ def reboot():
...
+      the internal flash, so destroying that secret is what makes them
+      unrecoverable. This is cryptographic erasure of the QSPI, not an
+      overwrite of it, and it is only as good as the internal-flash
+      overwrite that precedes it.
+    * Volatile memory is not scrubbed. The reboot below is
+      pyb.hard_reset() -> NVIC_SystemReset(), which does not clear SRAM,
+      and MicroPython's gc_init() only resets the allocation tables rather
+      than the heap itself. Secrets held in RAM can therefore survive the
+      reset in the physical SRAM until they are overwritten by ordinary
+      allocation.
```
maggo83, line 1083: @miketlk please check if this design is sufficient for specter security standards

src/platform.py
```diff
@@ -357,32 +978,156 @@ def reboot():
...
     """
     # delete files normally in simulator
     try:
```
maggo83, line 1086: if we assume this to pass only in simulator: why don't we query the simulator flag explicitly instead of this implicit approach?

src/platform.py
maggo83: #386 is followed up/replaced by by #396

maggo83 (dismissed):

maggo83 (commented):
new minor comments added

schnuartz-ai:
Good catch, and you're right — that was a regression I introduced with the overwrite-before-replace commit, and it is the worse of the two failure modes. Fixed in `f0d3c3b`.

### The order is now transactional

`_replace_existing()` is gone; `_save_key_file(fullpath, replacing)` handles both cases, and the replacing one runs:

```
write the new ciphertext under a temporary name
    -> strict sync
    -> read it back and verify it decrypts to what we meant to save
    -> only now overwrite and unlink the old file
    -> rename the temporary file into its place
    -> sync
```

Every failure before the secure delete — pulled card, failed write, failed sync, mismatched read-back — leaves the existing file exactly as it was, and the temporary file is cleaned up. After the secure delete, a complete and verified copy is already on the medium and what remains is a single directory update, rather than a gap in which no copy exists at all.

The old file is still never truncated over, so its cluster chain is not freed unoverwritten — which was the point of the original change.

### The leftover temp file

A save interrupted between the temp write and the rename leaves an encrypted phrase behind under the temp name. Two things follow from that, both handled:

* The next save **overwrites** it rather than truncating over it — otherwise the temp file reintroduces exactly the free-space residue this is trying to avoid.
* Its name (`tmpsave`) deliberately does not carry the `reckless` / `specterdiy<id>` prefix the key listings match on, so a leftover can never show up in the menus as a loadable key.

### `platform.sync()` swallowing sync errors

Also right, and also fixed for this path. `save_aead()` takes a `strict` flag:

```python
with open(path, "wb") as f:
    f.write(d)
    if strict:
        platform.strict_sync(f)
if not strict:
    platform.sync()
```

The recovery-phrase save is the caller that sets it. It stays **off** by default deliberately — most other `save_aead()` callers write state that is rebuilt on the next boot anyway, and `save_state()` in particular runs inside `_unlock()`, where any exception is converted into `CriticalErrorWipeImmediately`. Making the sync strict there would turn a transient sync hiccup into a device wipe, which is not a trade I want to make silently. If you'd rather have it strict everywhere, say so and I'll flip the default and deal with that path separately.

`_strict_file_sync()` is now the public `strict_sync(f=None)`, because the rename needs a filesystem sync with no file handle to flush. Its docstring keeps the caveat from the previous round: on the pinned MicroPython a clean return still is not proof the write reached the medium, since the VFS discards the block device's write status.

### What this does not fix

The remaining window — power loss between the secure delete and the rename — leaves a valid, verified `tmpsave` on the medium and no file under the expected name. The data is there but the UI will not offer it. Closing that properly needs a rename-into-place that is atomic against power loss, which FAT does not give us. If you want, the next step would be to keep the old file under a second name until the new one is in place (two directory entries, delete last) — that trades the window for a leftover the user might see. I'd rather do that as a follow-up than widen this PR further.

### Tests

`Ran 83 tests ... OK`. New coverage in `test_save_mnemonic_replace.py`:

- the new copy is written to the temp name and read back **before** any `secure_delete_file` of the target
- a failed write of the new copy, and a read-back that does not match, both leave the existing file byte-for-byte intact and remove the temp file
- a failed overwrite of the old file aborts the save with the old file intact
- a leftover temp file is securely deleted, not truncated over
- the temp name does not match the key-listing prefix
- the replacing save is the one that passes `strict=True`
- saving a genuinely new file still takes the plain path, with no temp file involved

schnuartz-ai:
Good catch, and I agree it was a merge blocker. Addressed in 3b58460, on top of f0d3c3b.

## Where it stands

f0d3c3b already moved to write-verify-then-replace, which fixes the data-loss window you described. 3b58460 closes what was left and fixes a regression the strict syncing introduced.

**1. The swap no longer destroys anything.** f0d3c3b still did `secure_delete_file(old)` and *then* `rename(tmp -> target)`. A cut between those two leaves the phrase stored only under the scratch name — which the pickers deliberately don't list, so the user sees an empty key list and has to be told about it in an error string.

The old file is now renamed aside instead:

```
write .tmp -> strict sync -> read back and verify
           -> rename target to .old
           -> rename .tmp onto target -> sync
           -> secure-delete .old
```

A FAT rename rewrites the directory entry and moves no data, so `.old` still occupies the clusters the target did and overwriting it at the end reaches exactly the same sectors — the anti-residue guarantee this PR is about is unchanged. But there is now no point in the sequence where zero complete copies exist under a name the code can find.

**2. The remaining gap is reconciled, not just reported.** The window between the two renames is recovered on the next save of that name, by `reconcile_scratch()`: restore `.old` if it is the only copy left, overwrite it if the swap already completed, always discard a `.tmp` (never verified, and it holds an encrypted phrase, so it is overwritten rather than unlinked). It runs *before* the "does this file exist" check, because the case that matters is exactly the one where the target is missing.

**3. Scratch names are per-target.** `.<name>.tmp` / `.<name>.old` rather than one shared `tmpsave` per directory, so a leftover belonging to one key can't be confused with — or destroyed by — a save of another. Both still start with a dot and match neither key-listing prefix.

**4. A failed write is a `KeyStoreError`, not a bare `OSError`,** which the device renders as a raw traceback (visible in the first screenshot below).

## A regression the strict syncing introduced

`strict_sync()` raised `OSError("no filesystem sync primitive is available")` when neither `os.sync` nor `os.fsync` exists, with a comment reading "MicroPython — the actual target — always has os.sync". That's true of the device firmware, which binds it in `ports/stm32/moduos.c`, but **not** of the unix simulator build, which binds neither:

```
>>> import os; [hasattr(os, n) for n in ("sync", "fsync")]
[False, False]
```

Once the save path started syncing strictly, that made **every** save of a recovery phrase fail on the simulator:

simulator regression

A runtime with no sync call is not a sync that failed. Where there's nothing to sync with, the flush is all there is to do; real errors from flush or sync still propagate, and the device behaviour is unchanged.

## On `platform.sync()` swallowing errors

Separated as you'd expect: `sync()` keeps its best-effort behaviour and now says so in its docstring, `strict_sync()` propagates, and `save_aead(strict=True)` checks for a short write and uses it. Opt-in, because files that are rewritten routinely shouldn't abort on a failed sync. The sync after close isn't a repeat of the one inside the `with` — closing is what writes the directory entry.

This doesn't make FAT transactional; the directory sectors are still updated without a journal, and as `strict_sync()` documents, a clean return isn't proof the data reached the card on this MicroPython. What it removes is the window where a single fault costs the user their only backup.

## Screenshots

Same scenario on both firmwares, driven through the simulator's TCP GUI with the same injected fault: a key `mykey` is already saved, the user saves over it and confirms the overwrite, and the first write of the new key file is rejected — as a pulled card or a failed SD write would reject it.

Both runs also get the same `os.sync` shim, so the simulator matches the hardware on that point; without it the old firmware's secure delete fails first and the comparison would show nothing.

**Before** — destroy-then-write, the version reviewed:

before

**After:**

after

On disk after the failed save, old vs new:

```
old:  enc_secret  pin  secret                      <- reckless.mykey is gone
new:  enc_secret  pin  secret  reckless.mykey      <- intact, no scratch files left
```

## Tests

The regression suite is reworked around the new contract, over both `FlashKeyStore` and `SDKeyStore`: the replaced copy is overwritten rather than unlinked; the replacement is already in place when that happens; a failed write, a replacement that doesn't read back, and a failed rename each leave the existing key untouched; a failed *retire* still leaves the key saved; each of the three power-cut leftovers is reconciled correctly; and scratch names are invisible to the pickers. Plus one for the missing-sync-primitive case above.

Full native suite: 92 tests, green.

Schnuartz:
Will proabably close and proceed with:
https://github.com/cryptoadvance/specter-diy/pull/408

Issue #229, opened 2022-11-15
Cutting power while verifying PIN code wipes the device.


stepansnigirev:
As the device is writing to the filesystem during PIN verification, cutting off the power can cause damage of the flash FS.

It can be mitigated if we split the flash to 2 parts, write data to the second part, and copy it to the first part when we are done.

This approach can be abused for an attack though - if the attacker bruteforces the PIN and cuts the power every time PIN verification fails he could rewind the device to the previous state. So we have to be very-very careful when implementing this.

Probably copy back and forth should be done in a few iterations to make sure we don't open an opportunity for such an attack.

schnuartz-ai:
# Draft comment for #229 — Cutting power while verifying PIN code wipes the device

## Current assessment

**STILL RELEVANT.** The issue is not fixed on HEAD (`3f3c831`). The flash-based PIN counter is written to the filesystem **before** the PIN is verified, and the pin-file load on boot wipes the device on any decryption/HMAC failure. A power loss that interrupts the in-progress write can leave the PIN file corrupt, which on the next boot triggers an unrecoverable wipe. The separately-filed #328 reports the same class of failure on the current firmware and confirms it is open and unresolved.

## What I investigated

- `src/keystore/flash.py` — `FlashKeyStore._unlock`, `save_state`, `load_state`, `create_empty_pin_file`.
- `src/keystore/ram.py` — `RAMKeyStore.save_aead` (the write+sync path), `wipe`, `unlock`.
- `src/platform.py` — `CriticalErrorWipeImmediately`, `sync`, `delete_recursively`, `wipe`.
- `src/specter.py` — `setup()` boot flow: `keystore.init()` → `load_state()` then `unlock()` → `_unlock()`.
- Git history: `git log -- src/keystore/flash.py`, `git show ec583cb` ("fix PIN verification bug"), `git log --grep="power|wipe|power loss"`.
- Related issues: #328 (open), #44.

## Current code

`FlashKeyStore._unlock` decrements the attempt counter and persists it to flash **before** checking the PIN (`src/keystore/flash.py:109-140`):

```python
def _unlock(self, pin):
    # if anything goes wrong here - wipe
    try:
        # decrease the counter
        self._pin_attempts_left -= 1
        self.save_state()                      # <-- writes pin file to flash + os.sync()
        # check we have attempts
        if self._pin_attempts_left <= 0:
            self.wipe(self.path)
[...]

PR #380, opened 2026-08-19
Fix device wipe on final PIN attempt

### LINKED ISSUE #7: Device wipes before checking the final PIN attempt
## Problem

With internal storage enabled, `FlashKeyStore._unlock()` decrements and saves the PIN-attempt counter before checking the entered PIN. It then wipes the device immediately when the counter reaches zero.

This means that after nine wrong PIN entries the UI reports one attempt remaining, but the tenth entry wipes the device before the PIN is checked. The correct PIN on the tenth entry therefore cannot unlock the device. Users effectively get only nine usable attempts instead of ten.

## Expected behavior

- Nine wrong PINs leave one attempt.
- The correct PIN on the tenth attempt unlocks the device.
- A wrong PIN on the tenth attempt wipes the device.

## Proposed fix

Keep decrementing and persisting the counter before PIN verification to preserve protection against power-cut rewind attacks (see #229). Verify the PIN after decrementing, and wipe only when the final attempt was actually wrong.

The fix is prepared in PR #6.

## Testing

Hardware testing is required before merging. The regression test should verify both the correct-PIN-on-attempt-10 and wrong-PIN-on-attempt-10 paths.

schnuartz-ai:
## Summary

Fix the internal-storage PIN attempt off-by-one behavior. Fixes Schnuartz/specter-diy#7.

`FlashKeyStore._unlock()` previously wiped the device as soon as the counter reached zero, before checking the entered PIN. After nine wrong entries, the UI reported one attempt remaining, but the tenth entry wiped the device even when the PIN was correct.

This change keeps decrementing and persisting the counter before verification to preserve the power-cut rewind protection discussed in #229. It verifies the PIN first, and wipes only when the final attempt was actually wrong.

## Expected behavior

- Nine wrong PINs leave one attempt.
- The correct PIN on the tenth attempt unlocks the device.
- A wrong PIN on the tenth attempt wipes the device.

## Testing

- `python -m py_compile src/keystore/flash.py`
- Logic simulation confirms the old and new tenth-attempt behavior.
- Hardware testing on a Specter DIY device is required before merging.


Schnuartz:
I tested that on Real Hardware.

The bug is real. In both cases I entered the wrong PIN nine times in a row, followed by one correct entry.

After entering it correctly:

In the current firmware, the device was wiped. In the fix I did, everything stayed the same, as it should.
IMG_20260819_140336

after that, I inserted the wrong PIN code ten times in a row, and this time, after the tenth time, everything got deleted as intended. 
Timelock receovery (3)

Schnuartz (approved):
Checked it on Hardware and had a look at the very tiny code. Approved.

Schnuartz:
@al-munazzim can you review

al-munazzim (approved):
Reviewed the final PIN-attempt change and ran local checks on the PR branch: `python3 -m py_compile src/keystore/flash.py` and `python3 test/run_native_tests.py` both completed successfully (native suite: 5/5). The counter is still decremented/persisted before verification, preserving the power-cut rewind protection, but wiping is now gated on an actually-wrong final PIN. Combined with the reported hardware test for correct-vs-wrong tenth attempt, this looks good to me.

maggo83 (approved):

PR #381, opened 2026-08-21
Require on-device confirmation for host XPUB requests

schnuartz-ai:
## Summary

`XpubApp.process_host_command()` returned the master fingerprint or an
XPUB at any host-selected derivation path immediately once the device was
unlocked, with no trusted-screen confirmation. Any process on the
connected computer that can reach the serial port could silently
enumerate the wallet fingerprint and every XPUB, and from those the
wallet's addresses and balance.

- **`fingerprint` stays non-interactive** — companion software (Specter
  Desktop, Liana) uses it for passive device discovery/identification.
  Still gated by the unchanged locked-device check.
- **`xpub <path>` now requires an on-device Confirm/Cancel**, showing the
  normalized derivation path and the exact XPUB that would be returned.
  Reject returns no key material (`error: User cancelled`).
- **`xpubauth begin <scope>` / `xpubauth end`** — one on-device
  confirmation authorizes a **bounded, explicitly enumerated** set of
  derivation paths (e.g. the four standard account keys, or a
  BIP-138-style discovery scope). Afterwards `xpub <path>` requests that
  fall inside the approved scope are answered without a further prompt,
  **each authorized path exactly once**; anything outside keeps prompting
  individually. The authorization is RAM-only, network-bound, and cleared
  on `xpubauth end` / lock / re-init / network change / self-exhaustion /
  a new `begin`. See `docs/communication.md` for the full grammar and
  lifecycle. This keeps the confirm-by-default privacy property while
  making multi-key import practical (N prompts → 1).
- **Wrong-network requests are refused.** A standard-purpose path
  (44'/48'/49'/84'/86'/87') whose coin type doesn't match the network the
  device is on can never be legitimately shared from there — the same
  rule an `xpubauth` scope is held to, also enforced for individual
  requests. The device shows a screen naming the requested derivation,
  the active network, and by name the network to switch to (`0'` →
  Mainnet, `1776'` → Liquid, `1'` → "Testnet, Signet or Regtest"). The
  host gets `error: network mismatch: device is on <network>`.
- **Input is bounded.** The `xpub` / `xpubauth` payloads are capped
  (`MAX_XPUB_PATH_LEN` / `MAX_SCOPE_COMMAND_LEN`) as they're read, before
  `.decode()`, and the parsed path depth is capped (`MAX_PATH_DEPTH`)
  before any BIP32 derivation. Path parsing catches only
  `(ValueError, IndexError)` instead of a bare `except`.
- `hwidevice.py` (this repo's bundled reference HWI client) drops the 3s
  timeout on `get_pubkey_at_path()` (a human confirms now) and gains
  `authorize_xpubs()` / `end_xpub_authorization()`. `get_master_fingerprint`
  keeps its short timeout. Signing is untouched.

## Companion-repo compatibility

- **Specter Desktop:** cryptoadvance/specter-desktop#2702 — matching
  timeout removal, wires up `xpubauth begin/end` for a single
  confirmation across the standard key set, and turns the `network
  mismatch` error into a "switch the device to X" message.
- **async-hwi:** wizardsardine/async-hwi#139 — maps a cancelled `xpub`
  request to `UserCancelled`/`UserRefused`.
- Older companion software is unaffected — it keeps sending individual
  `xpub <path>` requests, each confirmed on its own.

## Test plan

- [x] `test/run_native_tests.py` — 121/121, incl.
  `test_xpubs_host_command.py`, `test_xpub_scope.py`,
  `test_xpubauth_host_command.py` (scope grammar, range/overlap
  rejection, network binding, single-use consumption, self-exhaustion,
  fail-closed on replacement `begin`, cancel leaves no authorization,
  wrong-network refusal).
- [x] `make disco` builds clean (CI "Build Firmware").
- [x] Real STM32F469 Discovery hardware, over USB-serial and end-to-end
  through Specter Desktop's "Get via USB": one `xpubauth begin`
  confirmation, then the standard keys returned without further prompts;
  device on Mainnet / Regtest / Liquid — matching-coin-type request
  returns the key, wrong-network request shows the refusal screen and
  returns `error: network mismatch: device is on <net>`.

### DIFF
--- a/docs/communication.md
+++ b/docs/communication.md
@@ -61,11 +61,76 @@ We use human-readable plain text messages, because we can and they are way easie
 
 The following commands are supported:
 
-- `fingerprint` - returns hex fingerprint of the root key.
-- `xpub <derivation>` - returns xpub with derivation. For hardened derivation both `h` and `'` can be used. For example `xpub m/84h/1h/0h`.
+- `fingerprint` - returns hex fingerprint of the root key. Non-interactive: doesn't require on-device confirmation, so companion software can use it for passive device discovery/identification.
+- `xpub <derivation>` - returns xpub with derivation. For hardened derivation both `h` and `'` can be used. For example `xpub m/84h/1h/0h`. Unlike `fingerprint`, this **requires on-device confirmation** by default (see "XPUB privacy" below) - the device shows the requested path and the resulting xpub and waits for Confirm/Cancel. A cancelled request returns `error: User cancelled` without disclosing the xpub.
+- `xpubauth begin <scope>` / `xpubauth end` - scoped multi-XPUB authorization, letting a host request one bounded set of paths and then retrieve all of them via plain `xpub <derivation>` calls with a single confirmation. See "Scoped multi-XPUB authorization" below.
 - `sign <psbt>` - asks user to confirm transaction signing.
 - `showaddr <type> <derivation> [witness_script_hex]` - show address of `type` with `derivation`. `type` can be `wpkh`, `sh-wpkh`, `pkh`, `sh`, `sh-wsh` or `wsh`. Witness script is required for non-pkh wallets.
 - `importwallet <wallet_name>&<descriptor>` - asks user to confirm adding new `wallet` with `descriptor`.
+
+### XPUB privacy
+
+Any host connected over USB, or any process on the connected computer that can reach the Specter's serial port, can otherwise ask for wallet public keys and enumerate the wallet's addresses and balance without the user noticing. Because of that, a bare `xpub <derivation>` request always requires a physical Confirm/Cancel on the trusted display, showing exactly the derivation path and the resulting xpub that would be shared. This is deliberately different from hardware wallets that export a fixed set of "standard" xpubs silently - Specter's threat model assumes *any* program on the connected computer, not just the wallet software the user intends to use, may be sending these requests.
+
+Requesting many xpubs this way (e.g. wallet/account discovery, or [BIP-0138](https://github.com/bitcoin/bips/blob/master/bip-0138.mediawiki)-style recovery) would mean one confirmation per key, which is impractical. `xpubauth` (below) solves that without giving up the confirm-by-default privacy property.
+
+### Scoped multi-XPUB authorization
+
+`xpubauth begin <scope>` asks the user to approve, once, a bounded, explicitly enumerated set of derivation paths. After approval, `xpub <derivation>` requests that fall inside the approved scope are answered immediately, without another prompt; anything outside the scope keeps requiring the normal individual confirmation described above. `xpubauth end` discards the authorization explicitly. The host does not have to retrieve every path in the scope - stopping early (e.g. as soon as BIP-138 recovery succeeds) is fine.
+
+This is a logical, per-device-session permission, not tied to any particular transport connection - `xpub` requests over USB/simulator each open and close their own connection, so the authorization is tracked by the firmware itself, in RAM, independently of how many separate connections the host makes.
+
+#### Scope grammar
+
+```
+scope     := entry (";" entry)*
+entry     := "m" ("/" component)*
+component := index | range
+index     := digits ["h" | "H" | "'"]
+range     := "{" digits "-" digits "}" ["h" | "H" | "'"]
+```
+
+- Each entry is either an **exact path** (`m/84h/0h/0h`) or a path with **exactly one bounded range component** (`m/84h/0h/{0-9}h`, `m/48h/0h/{0-9}h/2h`). At most one range per entry.
+- A range's two numbers are inclusive, non-negative, non-hardened child indices; a trailing hardening marker (if present) applies to the whole range. There is no open range, no negative range and no wildcard (`*`) of any kind - every scope denotes a finite, precomputable set of paths, and the firmware rejects anything that isn't before displaying or allocating anything for it.
+- Multiple entries are separated by `;`, e.g. `m/84h/0h/{0-9}h;m/86h/0h/{0-9}h`.
+- Both `h`/`H` and `'` are accepted for hardened components and are equivalent - the device parses every path (both scope entries and later `xpub` requests) into normalized BIP32 integers and matches on that normalized form, never on raw text.
+- Duplicate or overlapping entries (e.g. `m/84h/0h/{0-9}h` together with `m/84h/0h/5h`) make the whole `begin` request invalid - overlap is rejected rather than silently merged, so the displayed and enforced count are always identical.
+- For recognizable standard purposes (`44h`, `48h`, `49h`, `84h`, `86h`, `87h`), the coin-type component must be a fixed value matching the network active on the device at the time of the request (see "Network binding" below); a scope mixing e.g. a mainnet and a testnet-style entry is rejected as a whole. Non-standard/custom paths carry no such implied network semantics and are left alone.
+- Hard limits (named constants in `apps/xpubs/scope.py`): at most 16 entries, at most 8 path components per entry, at most 200 values per range, at most 200 total xpubs per authorization, at most 1024 bytes for the whole `begin` argument. These comfortably cover the ~70-path BIP-138 discovery scope with headroom while keeping worst-case parsing, display and memory use trivial on the STM32F469.
+
+Example - the approximate BIP-138 mainnet discovery scope:
+
+```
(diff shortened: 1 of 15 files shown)


pythcoiner:
FYI this will make specter devices painful to use with [BIP-0138](https://github.com/bitcoin/bips/pull/1951)

I advise to enforce this only for uncommon derivation paths, as it's already done by Ledger/Bitbox devices

Schnuartz:
> FYI this will make specter devices painful to use with [BIP-0138](https://github.com/bitcoin/bips/pull/1951)
> 
> I advise to enforce this only for uncommon derivation paths, as it's already done by Ledger/Bitbox devices

Good idea. I think I will implement a extra function. So that it can ask for the permission to get multiple XPubs at once.

The difference to BitBox etc is that they currenly have a different Security Model. As they are building a channel with the companion App. Specter currently does not. Every program on the PC can ask for this information.

pythcoiner:
> Specter currently does not. Every program on the PC can ask for this information.

same as ledger/coldcard/jade (and Bitbox if the pairing metadata is persisted in clear on disk...)

> So that it can ask for the permission to get multiple XPubs at once.

then you should add a command for batch export of xpub to keep it usable, but then users will just blind ack imo...

Schnuartz:
@al-munazzim please review

al-munazzim:
Thanks for the ping. I did a code-path review around the host-command changes and ran the native test suite locally.

What I checked:
- `fingerprint` remains non-interactive but still respects the locked-device guard.
- bare `xpub <path>` now prompts after parsing/normalizing the path and shows both the normalized derivation and the actual XPUB that would be returned.
- `xpubauth` scopes are bounded, normalized before display/enforcement, network-checked for standard purposes, overlap-rejected, single-use per authorized path, and cleared on lock/re-init/network change/end.
- local native tests pass: `Ran 89 tests in 1.422s; OK` with `python test/run_native_tests.py` in a fresh uv venv.

No code-level blocker from me based on that review. The only thing I’d keep as an explicit merge/release consideration is companion compatibility/coordination: Specter Desktop needs cryptoadvance/specter-desktop#2702 (no short timeout for pubkey fetch), and async-hwi/Liana should handle user cancellation as described. Once those companion-side changes are accepted/released, this approach looks reasonable to me.

Schnuartz (approved):

schnuartz-ai:
### Hardware test — individual-request network guard + wording

Tested on a physical STM32F469 Discovery (this branch's `disco` build,
firmware `2041b8e`), both over raw USB-serial and end-to-end through
Specter Desktop's "Get via USB" flow.

**Wrong-network `xpub` request is refused, not prompted** — device on
Testnet, host asked for `m/49h/0h/0h` (mainnet coin type). The device
showed:

> **Host tried to get access to the following Xpub**
> Derivation: m/49h/0h/0h
>
> This device is currently on Testnet.
> This Xpub cannot be shared from here.
>
> To share it, switch the device to Mainnet in the settings first.
>
> `[ OK ]`

Single OK button — it never falls through to the normal "Share Xpub?"
Confirm/Cancel. On OK the host received `error: network mismatch: device
is on test`.

- `m/84h/1776h/0h` (Liquid coin type) → same screen, target named
  "Liquid", host gets `error: network mismatch: device is on test`.
- Repeated with the device set to **Mainnet**, **Regtest** and
  **Liquid**: in each case the matching-coin-type request goes through
  the normal confirmation and returns the key; the other two are refused
  with the correct target network named ("Mainnet" / "Testnet, Signet or
  Regtest" / "Liquid").

**`xpubauth` scope confirmation** — device on Mainnet, "Get via USB" for
a full account:

> **Share multiple XPUBs?**
> Device fingerprint ca73ce3b
>
> Network: Mainnet
> Allowed:
> m/49h/0h/0h
> m/84h/0h/0h
> m/48h/0h/0h/1h
> m/48h/0h/0h/2h
>
> Up to 4 extended public keys. Private keys are not shared.
>
> `[ Cancel ]   [ Allow ]`

One confirmation; the four subsequent `xpub` reads returned with no
further prompt.

`test/run_native_tests.py` — 118/118. `make disco` clean. CI green on
`2041b8e`.

---

**Device on Testnet, host asked for `m/49h/0h/0h` (via Specter Desktop "Get via USB"):**

wrong-network refusal screen + Specter Desktop requesting the key

**Device on Mainnet, same flow — one scope confirmation for the whole standard set:**

Share multiple XPUBs? confirmation on device

Schnuartz:
Tested it using real Hardware. Everything works
IMG_20260901_221808
IMG_20260901_221845
IMG_20260901_221921

pythcoiner:
> Share multiple XPUBs?
> Device fingerprint ca73ce3b
> Network: Mainnet
> Allowed:
> m/49h/0h/0h
> m/84h/0h/0h
> m/48h/0h/0h/1h
> m/48h/0h/0h/2h
> Up to 4 extended public keys. Private keys are not shared.
> [ Cancel ]   [ Allow ]

> One confirmation; the four subsequent xpub reads returned with no
further prompt.

implementations following BIP-0138 will have to fetch 70 xpubs in a block to workaround this new "feature", do you have tested it?

> All implementations that can fetch xpubs from a signing device SHOULD try the following
common paths automatically:

- `m/44h/<coin>h/{0..9}h`
- `m/49h/<coin>h/{0..9}h`
- `m/84h/<coin>h/{0..9}h`
- `m/86h/<coin>h/{0..9}h`
- `m/87h/<coin>h/{0..9}h`
- `m/48h/<coin>h/{0..9}h/1h`
- `m/48h/<coin>h/{0..9}h/2h`

schnuartz-ai:
### Reworked: `xpubauth` batch authorization removed

Force-pushed a rebuild of this branch. The PR is now a single commit on
top of master and drops the scoped/batch `xpubauth` protocol entirely.

**Gone:** `xpubauth begin` / `xpubauth end`, `src/apps/xpubs/scope.py`
(scope grammar, range/bitset tracking, `Authorization`, all the
`MAX_SCOPE_*` / `MAX_ENTRIES` / `MAX_TOTAL_XPUBS` limits), the RAM-held
`_authorization` state and its reset-on-lock / reset-on-network-change /
reset-on-init hooks, the `on_lock` app hook, `authorize_xpubs()` /
`end_xpub_authorization()` in `hwidevice.py`, the protocol docs, and the
`test_xpub_scope.py` / `test_xpubauth_host_command.py` /
`test/integration/tests/test_xpubauth.py` suites.

**Kept:** every `xpub <path>` export requires its own on-device
Confirm/Cancel (one confirmation authorizes exactly the key it's shown
for); `fingerprint` stays non-interactive; the wrong-network refusal
(screen + `error: network mismatch: device is on <net>`); the input
bound (`MAX_XPUB_PATH_LEN`) and path-depth cap (`MAX_PATH_DEPTH`); the
narrowed `except` around path parsing; the `hwidevice.py` timeout
removal.

Rationale: multiple confirmations for multiple keys are a deliberate
privacy property here, not a UX cost to optimize away. A batch/scoped
authorization — even RAM-only, network-bound and one-shot — is still
device-side state that a later request (or another process on the serial
port) could ride on between `begin` and `end`. Removing it makes the
guarantee simple: one physical confirmation ⇒ one xpub.

CI is green on the rebuilt head (Build Firmware + tests + native-tests).
Native suite 32/32.

@al-munazzim — the earlier code-path review still applies to the
confirm-each-export and wrong-network parts; the `xpubauth` scope
sections are moot now.

schnuartz-ai:
### Reverted the `xpubauth` removal — batch authorization is back

An earlier force-push had stripped the scoped/batch `xpubauth`
authorization from this PR. That was a mistake on my side and has been
undone: the branch is back to the state that has

- `xpubauth begin <scope>` / `xpubauth end` (and the `scope.py` grammar /
  `Authorization` / limits),
- one on-device confirmation covering a bounded standard key set (the
  "Share multiple XPUBs?" screen),
- single-use consumption, RAM-only state, cleared on lock / re-init /
  network change / self-exhaustion / replacement `begin`,

plus everything that was kept throughout: the per-request confirmation,
the wrong-network refusal + machine-parseable error, the 3s-timeout
removal, input bounding / depth cap.

The screenshots and older comments referencing "Share multiple XPUBs?"
apply again. Sorry for the churn.

Schnuartz (approved):
Tested on real Hardware (works very well if you tell AI to make some specific Tests. So AI Can connect to the board and send tests. 

Also did some with Specter Desktop fork.

Also reviewed code

Schnuartz:
@al-munazzim can you review

Schnuartz:
@al-munazzim can you review

Schnuartz:
https://github.com/cryptoadvance/specter-desktop/pull/2702
-> merged, so now we could merge in Specter DIY as well without causing problems

Issue #13, opened 2019-09-23
Allow to navigate between addresses


stepansnigirev:
In the "wallets" menu one should be able to display addresses and navigate through them

stepansnigirev:
Solved with https://github.com/cryptoadvance/specter-diy/pull/16

PR #383, opened 2026-08-21
test: add change-metadata attack vectors

schnuartz-ai:
## Summary

Adds three additional PSBT attack fixtures for the change-metadata protections
from PR #13. They use the same public `abandon ... about` seed as the original
demo, so a tester only needs to scan the included seed QR once.

Related PR: https://github.com/Schnuartz/specter-diy/pull/13

## Scenarios

1. **Forged Taproot change metadata**

   The host supplies Taproot BIP32 metadata claiming `/1/7` change, but the
   actual Taproot script does not match the local descriptor. The output stays
   visible with the invalid change metadata warning and its real address.

2. **Two forged change outputs**

   Two outputs claim branch-1 change with mismatching scripts. Both remain
   visible and each keeps its own warning, proving warnings are not overwritten
   or attached to the wrong output.

3. **Unknown input with genuine change**

   The output is a genuine `/1/10` wallet script, but input derivation metadata
   is absent. The existing unknown-wallet confirmation is shown and the output
   remains visible instead of being hidden as change.

## Files

- `0_seed_qr_abandon_about.svg` is the reusable seed QR from the original demo.
- Each numbered `.svg` is the scannable PSBT QR.
- Each numbered `.psbt` is the matching base64 fixture for USB or automation.
- `README.md` documents the exact values, addresses, claims, and expected UI.

## Verification

All three fixtures were run through the production
`WalletManager.preprocess_psbt()` path on PR #13 head `33684ba`:

- Scenario 1: mismatch warning on output 1, `change=False`.
- Scenario 2: independent mismatch warnings on outputs 0 and 1,
  `change=False` for all outputs.
- Scenario 3: unknown input wallet, genuine branch-1 output remains visible,
  `change=False`.

The native PR regression suite passed: 49 tests.

The full MicroPython Unix simulator requires `make unix`; `make` is not
available on the Windows preparation host.


schnuartz-ai:
Superseded by the clean fixture-only PR #384, which is based directly on upstream master and contains only the three attack-vector demos, QR codes, seed QR, and README.

Issue #12, opened 2019-09-23
Implement address verification


stepansnigirev:
Either bruteforcing or using format like `bitcoin:<addr>?index=<derivation index>`

stepansnigirev:
Solved with https://github.com/cryptoadvance/specter-diy/pull/16

PR #386, opened 2026-08-23
Fix incomplete secure wipe of QSPI flash

schnuartz-ai:
## Problem

`platform.wipe()` deleted the QSPI filesystem but physically overwrote only the internal-flash range. Most of the 16 MiB QSPI user-data region could therefore remain recoverable from the chip.

## Resolution

- Replace the hardcoded range with the verified internal-flash and full-QSPI block map.
- Overwrite both regions in bounded chunks without loading the QSPI region into RAM.
- Validate flash geometry before destructive work and explicitly flush write-behind caches.
- Track write, sync, unmount, and wipe failures; never treat a partial wipe as successful.
- Build both `USE_DBOOT=0` and `USE_DBOOT=1` firmware configurations in CI.

## Tests

The port includes native regression coverage for block boundaries, chunking, failures, geometry checks, sync ordering, and simulator behavior. CI also builds both firmware configurations.

This change was originally developed and tested in [Schnuartz/specter-diy#12](https://github.com/Schnuartz/specter-diy/pull/12). The original PR contains the full development history, test results, review discussion, and implementation details. The results below refer to the original source revision. Fresh CI on this upstream PR validates the cleanly ported revision against the current upstream master.

- `python3 test/run_native_tests.py`: 31/31 passed
- Both firmware configurations (`USE_DBOOT=0` and `USE_DBOOT=1`) were built/tested on the source branch

The original #12 source revision was not tested on real hardware for completion of the full 16 MiB QSPI wipe. No hardware validation is claimed for this clean upstream revision. The pinned MicroPython drivers also do not expose every possible underlying erase/program failure to Python; this limitation is documented in the implementation.

### DIFF
--- a/src/platform.py
+++ b/src/platform.py
@@ -2,7 +2,6 @@
 import sys
 import os
 import pyb
-import gc
 
 simulator = (sys.platform in ["linux", "darwin"])
 
@@ -392,10 +391,13 @@ def reboot():
 # remapping or translation layer sits between the block numbers above and
 # the physical sectors (see spi_bdev_writeblocks()/mp_spiflash_cached_write()
 # in ports/stm32/spibdev.c and drivers/memory/spiflash.c), so overwriting
-# every block in a range does physically destroy the data that was there.
-# Required sector-erase-before-write is handled transparently by that
-# driver. Sources checked: diybitcoinhardware/f469-disco (submodule pin
-# db3ce3e), diybitcoinhardware/micropython (submodule pin 6bdf1b6),
+# every block in a range overwrites all addressable QSPI sectors, which
+# prevents recovery through normal digital reads of the flash contents
+# (this says nothing about charge-remanence/decapping-level analog
+# recovery, which is out of scope for this threat model). Required
+# sector-erase-before-write is handled transparently by that driver.
+# Sources checked: diybitcoinhardware/f469-disco (submodule pin db3ce3e),
+# diybitcoinhardware/micropython (submodule pin 6bdf1b6),
 # ports/stm32/{storage.h,storage.c,flashbdev.c,spibdev.c},
 # ports/stm32/boards/STM32F469DISC/{mpconfigboard.h,mpconfigboard.mk},
 # ports/stm32/boards/stm32f469xi_dboot.ld, drivers/memory/spiflash.c.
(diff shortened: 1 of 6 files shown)


Schnuartz:
@al-munazzim can you review?

al-munazzim (approved):
Reviewed the patch and the added native wipe coverage. I also checked out the PR locally and ran `python test/run_native_tests.py` under a fresh venv; it passed (31 tests). CI is green as well.\n\nNo objections from my side. Real-hardware validation of the full QSPI wipe is still the important final confidence check before release, as the PR notes.

src/platform.py
```diff
@@ -357,32 +356,249 @@ def reboot():
...
+# the physical sectors (see spi_bdev_writeblocks()/mp_spiflash_cached_write()
+# in ports/stm32/spibdev.c and drivers/memory/spiflash.c), so overwriting
+# every block in a range overwrites all addressable QSPI sectors, which
+# prevents recovery through normal digital reads of the flash contents
+# (this says nothing about charge-remanence/decapping-level analog
+# recovery, which is out of scope for this threat model). Required
+# sector-erase-before-write is handled transparently by that driver.
+# Sources checked: diybitcoinhardware/f469-disco (submodule pin db3ce3e),
+# diybitcoinhardware/micropython (submodule pin 6bdf1b6),
+# ports/stm32/{storage.h,storage.c,flashbdev.c,spibdev.c},
```
maggo83: very long comments -> should probably be compacted or put elsewhere.

fragile: comments reference paths to files in subrepos -> might outdate/break silently

src/platform.py
```diff
@@ -357,32 +356,249 @@ def reboot():
...
+# ports/stm32/boards/STM32F469DISC/{mpconfigboard.h,mpconfigboard.mk},
+# ports/stm32/boards/stm32f469xi_dboot.ld, drivers/memory/spiflash.c.
+FLASH_BLOCK_SIZE = 512
```
maggo83, line 368: constants redefine information already available elsewhere in the project. Rather include/re-use instead of duplicating. Worst-case: derive them from C headers during build.

src/platform.py
maggo83, line 1: Functionally the changes are needed and seem mostly correct. However the placement of the code parts seems not according to the existing architecture. E.g. wipe() is appropriate, but its lower level helpers and constants should probably go directly into the f469disco repo (or be included from there).

The things that feel out of place are:

Hard-coded physical flash block ranges (0x100, 192, 32768).
References to linker-script regions (FLASH_FS, FLASH_TEXT, FLASH_BOOT1/2).
References to MicroPython internal files (flashbdev.c, spibdev.c, storage.h).
Detailed proof that the QSPI chip has no translation layer.
These are firmware/board-level knowledge, not application-level knowledge. They are specific to:

STM32F469 Discovery
The USE_DBOOT=1 linker layout
The specific MicroPython fork and pin
The specific QSPI NOR chip
All of that lives in the f469-disco submodule.

Another approach that could help to get rid of some of the constants while staying correct is to query at run-time (e.g. from pyb.Flash():

f = pyb.Flash()
block_size = f.ioctl(5, None)
block_count = f.ioctl(4, None)

src/platform.py
```diff
@@ -357,32 +356,249 @@ def reboot():
...
+        # change ever shifted them, that mismatch could otherwise go
+        # unnoticed - the tests only check the code against itself, not
+        # against real hardware. Fail closed instead of guessing: refuse
+        # to touch either filesystem rather than wipe based on assumed
+        # boundaries that may no longer hold.
+        f = pyb.Flash()
+        block_size = f.ioctl(5, None)
+        block_count = f.ioctl(MP_BLOCKDEV_IOCTL_BLOCK_COUNT, None)
+        if block_size != FLASH_BLOCK_SIZE or block_count != QSPI_END_BLOCK:
+            raise RuntimeError(
```
maggo83, line 505: I assume we call wipe() on possibly infiltrated devices. Should we not try to wipe as much as possible here, even if we encounter unexpected sizes? If the attacker can control ioctl he could prevent the wipe this way.

src/platform.py
```diff
@@ -357,32 +356,249 @@ def reboot():
...
+    reused (sliced for the final, possibly shorter chunk) instead of
+    allocating fresh memory every iteration.
+
+    A failed chunk does not stop the wipe: we keep going so as much of
+    the remaining sensitive data as possible still gets destroyed. But
+    the return value reflects whether *every* chunk succeeded - callers
+    must not treat a False result as if the region were fully wiped.
+    """
+    zeros = bytes(chunk_blocks * block_size)
+    ok = True
```
maggo83, line 461: Isn't it more defensive to start with assumed failure and only assume ok after all blocks have demonstrably been written?

src/platform.py
schnuartz-ai: Fixed in b93cdca. The long derivation moved to docs/flash-block-map.md, the code keeps a short block map and a pointer to it. The references into the submodules are in the doc now as well, with a note to re-verify them when a submodule is bumped. So if they go stale it happens in a doc and not in the middle of the wipe code.

src/platform.py
schnuartz-ai: You are right, this is board level knowledge and not application level. I opened diybitcoinhardware/f469-disco#44 to get this exposed as a proper API there. Until that exists I kept the split point here, because there is nothing to import from yet. Once it lands I will switch to it.

About your ioctl suggestion: that is done now in cd1012a. Block size and block count are read back from the device and used directly, they are not constants anymore. The only value left is the split point between /flash and /qspi, which no ioctl exposes.

src/platform.py
schnuartz-ai: Changed in cd1012a. Block size and the end of the QSPI region are not constants anymore, wipe() reads both back from the block device with ioctl(). What is left is the split point between /flash and /qspi, because no ioctl exposes it. That one comes from the linker script and storage.h in the submodule, so I opened diybitcoinhardware/f469-disco#44 to get it queryable. Once that exists the last constant can go too.

src/platform.py
schnuartz-ai: I thought about this one. If an attacker can control what ioctl returns, he already runs code on the device. Then he can patch out the wipe call directly, he does not need the size mismatch for that. So I do not think the check gives him anything he does not have already.

What the check does protect against is the other case: a new board revision or a MicroPython bump shifts the block numbers, we write to the wrong range, report success, and the seed is still there. That failure is silent and I find it more likely than the attacker case.

Also the device does not stay usable when the check fails. wipe_or_halt() shows a warning and reboots either way, so nobody ends up with a device that looks wiped but is not.

That said, your point made me narrow it down. With cd1012a the wipe no longer refuses just because the chip size differs from the one we verified against. It wipes to whatever end the device reports, so a bigger or smaller chip gets wiped as far as it goes instead of not at all. It still refuses if the device does not even reach the assumed split point, because then we would be writing to block numbers that mean nothing.

src/platform.py
schnuartz-ai: I see the point and as a general rule I agree. In this loop I do not think it helps. Every failure path already sets ok = False, so the result is the same either way. And there is one case where starting from False would be worse: if start_block == end_block there is nothing to write, and we would report a failure although nothing failed. That is why I left it as it is. If you still prefer the other direction I can change it, it is not important to me.

schnuartz-ai:
Closed this by accident, and the branch deletion in the same action means GitHub will not let it be reopened — `gh pr reopen` and the REST endpoint both return 422, and the page shows no reopen control even after restoring the head branch to `b93cdca`.

Continued in #396, same branch, no changes made for the move. @maggo83 your two open points (the ioctl-controlling attacker, and the pessimistic `ok` flag) are carried over there and still unanswered from my side — I argued against both rather than changing the code, so please push back there if you disagree. @al-munazzim sorry for the churn, your approval does not carry over automatically.

PR #387, opened 2026-08-23
Fix change-output classification to require verified descriptor derivation

schnuartz-ai:
## Problem

Output metadata supplied by the host could cause a wallet-owned output to be treated as change without proving that it was derived from the wallet's change branch. This could hide a receive-branch or otherwise unverified output from the primary confirmation view.

## Resolution

- Require one unambiguous spending wallet, exactly two descriptor branches, branch-list position 1, and an on-device script re-derivation that matches the output script before accepting automatic change.
- Fail closed for ambiguous, unusual, or unverifiable cases.
- Keep unverified outputs visible and label them according to the actual `is_change` verdict.
- Surface forged change metadata and unverified internal outputs as warnings.
- Apply the same classification rules to Bitcoin and Liquid and add regression coverage for legacy/segwit and Taproot cases.

## Security invariant

An output may be automatically treated as change only after the device cryptographically verifies the matching wallet descriptor derivation. Host-supplied derivation metadata is never trusted by itself. Everything else remains visible.

## Tests

Regression tests cover receive versus change branches, forged metadata and script mismatches, unknown and mixed wallets, unusual branch values and branch counts, Taproot, Liquid, output labels, warning propagation, and confirmation-screen visibility.

This change was originally developed and tested in [Schnuartz/specter-diy#13](https://github.com/Schnuartz/specter-diy/pull/13). The original PR contains the full development history, test results, review discussion, screenshots, and reproduction material. The results below refer to the original source revision. Fresh CI on this upstream PR validates the cleanly ported revision against the current upstream master.

Extensive manual hardware validation was performed on the source PR, including the critical regression case, a multi-output case, an external-payment sanity check, genuine verified change, and forged branch-1 metadata with a script mismatch. See the [original hardware test comment](https://github.com/Schnuartz/specter-diy/pull/13#issuecomment-5359593356) and the [updated hardware results](https://github.com/Schnuartz/specter-diy/pull/13#issuecomment-5368643364).

The later corrected attack fixtures were also tested on hardware as valid testnet transactions: forged Taproot change metadata, two forged change outputs, and an unknown input with a genuine internal/change output. The tester documented the expected behavior and validity of all three transactions in this [direct test comment](https://github.com/Schnuartz/specter-diy/pull/13#issuecomment-5379167093).

The hardware results above refer to the original source revision, not this rebased upstream commit. Fresh hardware validation of this clean port remains recommended.


Schnuartz:
@al-munazzim can you review?

al-munazzim (approved):
Reviewed the change-classification fix and the expanded regression coverage. I also checked out the PR locally and ran `python test/run_native_tests.py` under the review venv; it passed (41 tests). CI is green.\n\nThe conservative rules here look right to me: branch-0/self-payment and ambiguous/unknown-input cases stay visible, verified branch-1 change is still recognized, and invalid host-supplied change metadata gets surfaced as a warning.

schnuartz-ai:
## Manual test vectors for PR #387

> ⚠️ PUBLIC DEMO SEED — Never use this mnemonic for real funds. Anyone can reconstruct and control this wallet.

Tested branch: `Schnuartz/specter-diy:codex/pr-387-change-metadata`  
Commit: `b2bbd19` (`b2bbd199081d37af31afde01dbfff06c4636b0b7`)

All transaction vectors below are regtest-only PSBTs. The QR images below are deterministic PNGs with an explicit white background (#ffffff) and black modules (#000000); scan one complete image at a time.

### Demo wallet setup

**Mnemonic**

`ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost machine`

**Standard SeedQR payload**

`078007800780078007800780078007800780078007801068`

**Scannable SeedQR**

SeedQR - white background

Start a fresh simulator/device state. From the top-level menu choose **Import recovery phrase**, select the QR host, and scan the SeedQR above. Confirm the displayed mnemonic and use an empty passphrase. This creates the default demo wallet.

### Test 1 — Forged change metadata / Script mismatch

Use the demo wallet from the setup above. From the signing/QR-host flow, scan the complete single QR below. Confirm the transaction and open the output details. The PSBT has one wallet-owned input, but its output metadata claims branch 1 index 9 while the actual output script is unrelated to the wallet.

**Decoded PSBT payload**

`cHNidP8BAFICAAAAARERERERERERERERERERERERERERERERERERERERERERAAAAAAD/////AaCGAQAAAAAAFgAU/k43miJj0GnfO4zXUv8BpknP0i8AAAAAAAEBH8DUAQAAAAAAFgAUYtpvUJWlKzQMGupTEzQTZK+8Lo4iBgMx7csWz9D4WYBS97KHsHBHoRxglnwefrAlfgJVJTnZhBiMJKUQVAAAgAEAAIAAAACAAAAAAAAAAAAAIgIDKphZWX45d13FRc7DcU5R4yt9mqRKLtOMAtD10rgLl7oYjCSlEFQAAIABAACAAAAAgAEAAAAJAAAAAA==`

**Scannable Test 1 QR**

Test 1 QR - white background

**Expected:** The output remains visible, is not classified as change, and this warning is shown: `Invalid change metadata! Host claimed this output as wallet change, but it does not match your wallet. Verify the destination.`

### Test 2 — Conflicting Branch-1 claims

Reset to the same clean demo-wallet state, then scan the complete single QR below in the signing/QR-host flow. The output metadata contains two different valid branch-1 claims (`/1/3` and `/1/4`) for the sole spending wallet, while the output script is unrelated. Confirm the transaction and inspect the output details.

**Decoded PSBT payload**

`cHNidP8BAFICAAAAARISEhISEhISEhISEhISEhISEhISEhISEhISEhISEhISAAAAAAD/////AaCGAQAAAAAAFgAU/k43miJj0GnfO4zXUv8BpknP0i8AAAAAAAEBH8DUAQAAAAAAFgAUYtpvUJWlKzQMGupTEzQTZK+8Lo4iBgMx7csWz9D4WYBS97KHsHBHoRxglnwefrAlfgJVJTnZhBiMJKUQVAAAgAEAAIAAAACAAAAAAAAAAAAAIgICO5YXWu4yAeGzKWuZs3VnYO7DHlURTZey3y+3kNLvg2EYjCSlEFQAAIABAACAAAAAgAEAAAADAAAAIgIDAR9Aq14nFXryqCEzJ1/krMFkJshsgpom7APwYVDu4DIYjCSlEFQAAIABAACAAAAAgAEAAAAEAAAAAA==`

**Scannable Test 2 QR**

Test 2 QR - white background

**Expected:** Fail closed; no unsafe descriptor re-derivation is used, the output is not hidden or classified as change, and the inconsistent metadata warning is shown.

### Test 3 — Three-branch descriptor `<0;1;2>`

Reset to a clean demo-wallet state. Import this descriptor as a second wallet named `Three branch` before scanning the PSBT. On the simulator, send the following exact line to the QR-host TCP port (`5941`) or import the same descriptor through the normal wallet-import channel, then confirm the wallet import:

`Three branch&wpkh([8c24a510/84h/1h/7h]tpubDC4DsqH5rqHqzT3i7tAPiwmYUC76zoyhPpgecpAuD4XhjBErAi92Hh3BxJsPJPGuciECGTTHgQgqyV3ThUvbkzfwP5zFkmvRLz4RXSbBrVp/{0,1,2}/*)`

The descriptor must be stored/imported so the wallet is controlled by the demo keystore. Then scan the complete single QR below in the signing/QR-host flow, confirm the transaction, and inspect the output details.

**Decoded PSBT payload**

`cHNidP8BAFICAAAAARMTExMTExMTExMTExMTExMTExMTExMTExMTExMTExMTAAAAAAD/////AaCGAQAAAAAAFgAUK9puvAse0UJDQKoPTNEAEuxesKEAAAAAAAEBH8DUAQAAAAAAFgAUvuNIW28GRhJ07i5dmtJ+V+Bm07oiBgMFZIhwGoWBLqh74aDNphmj6FP6FSe3LMj1aO9e8fAVDRiMJKUQVAAAgAEAAIAHAACAAAAAAAAAAAAAIgID90EjrTwE/errKReRkC/7MNzCck+mmdBcZ1Og7VvUuT0YjCSlEFQAAIABAACABwAAgAEAAAADAAAAAA==`

**Scannable Test 3 QR**

Test 3 QR - white background

**Expected:** The descriptor has three branches; branch 1 is not automatically treated as change, the output remains visible, and no incorrect change-metadata warning is shown. A normal wallet-control label may still identify `Three branch`.

### Verification results

- SeedQR decoded: exact 48-digit payload; eleven `ghost` indices (`0780`) and one `machine` index (`1068`); BIP39 checksum valid.
- Test QR decoding: 3/3 exact PSBT payloads decoded from the rendered QR images.
- Specter DIY QR host/parser and PSBT preprocessing path: 3/3. Each payload was accepted as `SIGN_PSBT`; Test 1 and Test 2 produced `change=False` plus the invalid-metadata warning; Test 3 produced `change=False` with no warning after importing the descriptor with the demo keystore.
- Native suite at `b2bbd19`: 41/41 passed.
- Firmware-on-device tests: not run in this environment.
- Bitcoin Core RPC tests: not run; no Bitcoin Core regtest node was available.

schnuartz-ai:
## Manual test result for PR #387

Okay, we tested it. The firmware was tested with all three transaction scenarios, and everything works as expected.

Tested branch: `Schnuartz/specter-diy:codex/pr-387-change-metadata`  
Commit: `b2bbd19`

### Test 1

The two photos marked with `1` belong to the first transaction. The expected forged-change metadata behavior is visible.

Test 1 - photo 1

Test 1 - photo 2

**Result:** Works as expected.

### Test 2

The two photos marked with `1 2` belong to the second transaction. The expected conflicting-claims behavior is visible.

Test 2 - photo 1

Test 2 - photo 2

**Result:** Works as expected.

### Test 3

The three photos marked with `1 2 3` belong to the third transaction. The three-branch descriptor behavior is visible.

Test 3 - photo 1

Test 3 - photo 2

Test 3 - photo 3

**Result:** Works as expected.

### Summary

All three manual test cases were completed successfully. The correct output visibility and warning behavior were observed on the tested firmware.

docs/security-model.md
```diff
@@ -304,10 +304,34 @@ The following rules apply to transactions that the wallet will sign:
...
+The Bitcoin transaction does not contain an `is_change` flag. A PSBT may carry
+host-supplied BIP32 or Taproot derivation metadata, but that metadata is
+untrusted and is checked against the locally stored descriptor and the actual
+output script. If a branch-1 wallet claim does not match, the output remains
+visible and the device displays: `Invalid change metadata! Host claimed this
+output as wallet change, but it does not match your wallet. Verify the
+destination.` This reports inconsistent host metadata; it does not claim that
+Bitcoin marked the output as change.
+
+The forged branch-1/script-mismatch demonstration sometimes called scenario
```
maggo83: is this paragraph stateless documentation or documenting historical evolution that belongs into the commit description/history instead of a permanent documentation of the current state of things?

src/apps/wallets/liquid/manager.py
```diff
@@ -542,31 +544,50 @@ def preprocess_psbt(self, stream, fout):
...
                     idx, branch_idx = res
-                    branch_txt = ""
-                    if branch_idx == 1:
-                        "change "
-                    elif branch_idx > 1:
-                        "branch %d " % branch_idx
-                    metaout["label"] = "%s %s#%d" % (wallet.name, branch_txt, idx)
+                    if is_change:
+                        # Verified change is hidden from the primary
+                        # confirmation screen (see TransactionScreen), so
```
maggo83: is this comment relevant here? it refers to the disply logic which is coded elsewhere I assume -> if that changes will that comment be obsolete?

src/apps/wallets/manager.py
```diff
@@ -385,6 +396,100 @@ async def confirm_wallets(self, wallets, show_screen):
...
+            ):
+                try:
+                    desc, _ = wallet.get_descriptor(idx, branch_idx)
+                    if desc.script_pubkey() == out.script_pubkey:
+                        warning = UNVERIFIED_CHANGE_WARNING % (branch_idx, idx)
+                except Exception:
+                    pass
+        return derivation if wallet is not None else None, is_change, warning
+
+    def get_verified_change_derivation(self, wallet, wallets, out):
```
maggo83: This seesm to be dead code. Even if it is called: any caller of get_output_state can easily drop the last output so this function is really overkill imo

src/apps/wallets/manager.py
maggo83, line 1: The output metadata update (change classification, warning handling, label formatting, gap-limit check, watch-only warning) is duplicated almost verbatim between WalletManager and LWalletManager. Since LWalletManager inherits from WalletManager and both use the same get_output_status / add_output_warning helpers, this logic should be extracted into a base-class method. Only the Liquid-specific parts (asset handling, skip_separator, rangeproof separator) should remain in liquid/manager.py.

src/gui/screens/transaction.py
```diff
@@ -65,11 +65,18 @@ def __init__(self, title, meta):
         self.style_warning = style_warning
         self.style_gray = style_gray
 
-        num_change_outputs = 0
         for out in meta["outputs"]:
-            # first only show destination addresses
+            # A verified change output (see
+            # WalletManager.get_verified_change_derivation - branch 1,
+            # on-device script re-derivation matches) doesn't need its own
```
maggo83: I really don't like those verbose cross referencing comments. Very easy to break in the future.

maggo83 (changes requested):

docs/security-model.md
schnuartz-ai: Agreed – that paragraph documented how the behaviour evolved rather than the current state. Removed in 0cfab5d. What stays in `security-model.md` is stateless: the exact conditions under which change is auto-verified (one unambiguous spending wallet, exactly two branches, branch-list position 1, on-device script re-derivation) and what the `Invalid change metadata!` warning means.

src/apps/wallets/liquid/manager.py
schnuartz-ai: Right – that comment described the `TransactionScreen` display logic living in another file and would have gone stale the moment that changed. In 0cfab5d the whole label/warning block was moved out of the Liquid override into the shared `WalletManager.fill_output_metadata()`, so the comment is gone entirely. `LWalletManager.preprocess_psbt()` now only fills in `value` / `asset` / `raw_asset` and calls the base method.

src/apps/wallets/manager.py
schnuartz-ai: Correct – it had no production caller. `preprocess_psbt()` runs `get_output_status()` directly. Removed `get_verified_change_derivation()` in 0cfab5d and pointed the regression tests at `get_output_status()` (the function the firmware actually executes).

src/apps/wallets/manager.py
schnuartz-ai: Done in 0cfab5d. The shared block – change flag, address, label formatting, gap-limit check and watch-only warning – is now `WalletManager.fill_output_metadata(metaout, wallet, wallets, out)`. `LWalletManager.preprocess_psbt()` sets only `value` / `asset` / `raw_asset` and calls the base method; the Liquid-specific parts (asset commitments, `skip_separator`, rangeproof/surjection separator) stay in `liquid/manager.py`. `get_output_status()` / `add_output_warning()` are unchanged and shared as before.

src/gui/screens/transaction.py
schnuartz-ai: Trimmed in 0cfab5d. The comment no longer references functions in other files – it is down to three lines explaining why a verified-change output with no warning is skipped in the primary confirmation loop (it still appears in full on the details page). The cross-reference to `WalletManager.get_verified_change_derivation` is gone (that function was removed too).

schnuartz-ai:
Follow-up: gated the forged-change-metadata warning on `num_branches == 2` (9e47211).

In the `wallet is None` path of `get_output_status()` the `INVALID_CHANGE_METADATA` warning fired for any descriptor-valid branch-list position-1 claim of the sole spending wallet, independent of branch count. That contradicted the rest of the PR: for a `<0;1;2>` descriptor, position 1 has no defined "change" meaning (the `is_change` path already requires `num_branches == 2`), yet a position-1 host derivation with a mismatching output script still produced *"Invalid change metadata! Host claimed this output as wallet change…"* — the host only supplied a derivation, not a change claim.

Not exploitable (the output stayed `change=False` and visible), but semantically wrong and inconsistent with `is_change` and the docs. Both paths now use the same `num_branches == 2` gate.

Regression test added (`test_three_branch_forged_position_one_metadata_does_not_warn`): `<0;1;2>` sole spending wallet + valid position-1 metadata + mismatching output script → `change=False`, **no** warning. Native suite: 42/42.

Schnuartz (approved):

src/apps/wallets/manager.py
```diff
@@ -385,6 +396,142 @@ async def confirm_wallets(self, wallets, show_screen):
...
         return proceed
 
+    def add_output_warning(self, metaout, warning):
+        """Append an output warning without discarding an earlier warning."""
+        if not warning:
+            return
+        warnings = metaout.setdefault("warnings", [])
+        if warning not in warnings:
+            warnings.append(warning)
+        metaout["warning"] = "\n".join(warnings)
```
maggo83: General question: is it smart to always simultaneously update the list of warnings and the resulting warning text/string? Isn't is safer, easier and more efficient to only manage the list of warnings and only construct the resulting warning text/string when needed?

src/apps/wallets/manager.py
```diff
@@ -385,6 +396,142 @@ async def confirm_wallets(self, wallets, show_screen):
...
+
+    def get_wallet_derivation_claims(self, wallet, out):
+        """Return all descriptor-valid derivation claims made for an output."""
+        claims = []
+        for derivation in getattr(out, "bip32_derivations", {}).values():
```
maggo83: here we do have two 99% identical loops. could we refactor bip32_derivations and taproot_bip32_derivations such that we can rather have an outer loop iterating over bip32/taproot_bip32 and then a common inner loop?

src/apps/wallets/manager.py
```diff
@@ -385,6 +396,142 @@ async def confirm_wallets(self, wallets, show_screen):
...
+    def get_output_status(self, wallet, wallets, out):
+        """Return ``(derivation, is_change, warning)`` for one output.
+
+        A failed ownership check must not erase a descriptor-valid claim from
+        the sole known spending wallet. Claims from unrelated imported wallets
+        and ambiguous transactions are deliberately ignored.
+        """
+        derivation = None
+        if wallet is not None:
+            derivation = wallet.get_derivation(
```
maggo83: Proposal for a refactor: change signature/footprint of get_derivation to accept one single input/output and then return the bip32/taproot descriptors either in a dict or as separate return values.
This way the getter/fallback logic like getattr(out, "taproot_bip32_derivations", {}") can be unified and only has to be maintained in one place.
We already have inconsistent behavior for this in the code base -> errorprone.
also the calls get simpler and shorter.

src/apps/wallets/manager.py
```diff
@@ -385,6 +396,142 @@ async def confirm_wallets(self, wallets, show_screen):
...
+                and None in wallets
+                and wallet.descriptor.num_branches == 2
+                and branch_idx == 1
+            ):
+                try:
+                    desc, _ = wallet.get_descriptor(idx, branch_idx)
+                    if desc.script_pubkey() == out.script_pubkey:
+                        warning = UNVERIFIED_CHANGE_WARNING % (branch_idx, idx)
+                except Exception:
+                    pass
```
maggo83, line 582: Not sure I fully understand yet how we handle/display send/change/fee amounts. But to me it seems this warning is not addressing the root cause/issue (=we have inputs from unknown wallets. That in itself could/should have a warning).

First understanding:
-count all inputs from "us"=wallets which have a contribution to our key(s) = (our) base_amount
-(our) return_amount = all inputs going to "us"=wallets we have registered
-fee_mount = all inputs - all outputs
-(our) send_amount = our base_amount - our return_amount

for display i would suggest following rules:
if all inputs are from us:
     - we definitely paid the fees, so for display, subtract fees from displayed send amount.
     -if all outputs are to us: this seems to be a consolidation or so: label as self send and don't display send_amount (as it would equal 0), but only fee and returned value
     -else: label return_amount as change and do not display it. Only show (send_amount-fees) and fees
if at least one input is from someone else:
     -it is unclear who paid the fees and not traceable how in/out relate to each other. display warning to user
     -display to the user: send_amount, return_amount and fees all separately

All descriptions above only go for the default simplified initial display to the user. upon request as full/more detailed breakdown should always be possible for the user.

summary from my AI:

| Situation | What to show | PR 387 behavior |
|---|---|---|
| All inputs ours, only our outputs | consolidation / self-send: fee + returned value | Uses `change` flag |
| All inputs ours, external outputs | (send_amount − fees) + fees, label return as change | Uses `change` flag |
| Unknown inputs present | **send_amount** (our outflow) + **return_amount** + **fees**, plus warning | Per-output “unverified change” warning, `send_amount` inflates |

src/apps/wallets/manager.py
schnuartz-ai: Implemented in `fbca920`. `metaout["warnings"]` is now the single source of truth; `add_output_warning()` only appends and deduplicates warnings. The UI joins the list at the transaction/details-page boundary, so the list and a cached warning string cannot diverge. The regression tests were updated accordingly.

src/apps/wallets/manager.py
schnuartz-ai: Implemented in `fbca920`. `get_wallet_derivation_claims()` now feeds the normal BIP32 values and a normalized Taproot derivation iterator through one shared inner loop. Taproot leaf metadata is discarded only at that boundary; descriptor validation and claim deduplication remain unchanged.

src/apps/wallets/manager.py
schnuartz-ai: Agreed. I left the public `get_derivation()` API unchanged in PR #387 so this security fix does not widen its regression surface. I opened [issue #409](https://github.com/cryptoadvance/specter-diy/issues/409) to refactor it to accept one PSBT input/output scope and centralize the BIP32/Taproot fallback handling, with coverage for both metadata forms.

src/apps/wallets/manager.py
schnuartz-ai: Agreed that mixed-input transactions need explicit accounting semantics. Specter DIY already keeps the existing unknown-input warning and asks the user to confirm before proceeding; PR #387 also keeps every output visible unless change was verified on-device. I kept the broader amount/fee presentation redesign out of this security fix and opened [issue #410](https://github.com/cryptoadvance/specter-diy/issues/410) for the separate mixed-input display model, including known-wallet contribution, return amount, external outputs, total fee, and the uncertainty of fee attribution.

maggo83 (approved):

schnuartz-ai:
I tested it with the Specter DIY simulator using the same seed and transaction.

This is after, and this is before.

| Before | After |
| --- | --- |
| Before - fbca920 | After - e2123e1 |
| [fbca920 — Address PR 387 review cleanup](https://github.com/cryptoadvance/specter-diy/commit/fbca920b64ea83295ba35de8fe69fb1a26be6651) | [e2123e1 — Improve validation of PSBT change claims](https://github.com/cryptoadvance/specter-diy/commit/e2123e1eff814c16f6e89cfd3e4e06e9f5470e91) |

### Reproducible demo data

Demo seed phrase (intentionally public):

`ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost machine`

SeedQR (white background):

Demo seed QR

Transaction PSBT QR (white background):

Demo transaction PSBT QR

Schnuartz (approved):
Checked again. Changes from Mike make sense.

Approved

maggo83 (approved):
I do think a follow-up cleanup to make the logic more transparent/simple and efficient might be added, but this is already a substantial improvement and see no hard blockers to merge it.

PR #388, opened 2026-08-23
Use constant-time comparison for security-sensitive MAC/PIN checks

schnuartz-ai:
## Problem

Security-sensitive MAC and PIN checks used ordinary byte equality. Such comparisons may stop at the first differing byte and can expose timing information.

## Resolution

Add a centralized fixed-work `consteq()` helper for fixed-length authentication values and use it for storage-encryption MACs, PIN verification, and all JavaCard SecureChannel HMAC checks. The existing cryptographic protocols, key derivation, and failure behavior are otherwise unchanged.

## Tests

The port includes regression coverage for `consteq()`, AEAD/MAC validation, PIN verification, and SecureChannel decrypt/open handshakes.

This change was originally developed and tested in [Schnuartz/specter-diy#14](https://github.com/Schnuartz/specter-diy/pull/14). The original PR contains the full development history, test results, review discussion, and implementation details. The results below refer to the original source revision. Fresh CI on this upstream PR validates the cleanly ported revision against the current upstream master.

- Native tests: 5/5 passed
- MicroPython `make test`: 28/28 passed
- AEAD/MAC tests, PIN verification, and SecureChannel tests passed
- Production firmware `make disco`: successful

Full original test history and implementation discussion: [Schnuartz/specter-diy#14](https://github.com/Schnuartz/specter-diy/pull/14)

### DIFF
--- a/src/helpers.py
+++ b/src/helpers.py
@@ -49,6 +49,12 @@ def consteq(a: bytes, b: bytes) -> bool:
     do not use it to compare secret-dependent variable-length data.
     Used because the MicroPython build on this device (custom C `hmac`
     usermod) has no `hmac.compare_digest()`.
+
+    SECURITY: do not "simplify" the loop below back to `a == b` / `!=`,
+    and do not add an early `return False` inside the loop - either
+    change reintroduces the timing side channel this function exists
+    to remove. A test asserting equal() == equal() cannot catch that
+    regression, since both implementations return the same booleans.
     """
     if len(a) != len(b):
         return False
@@ -115,6 +121,7 @@ def aead_decrypt(ciphertext: bytes, key: bytes) -> tuple:
 
     aes_key = tagged_hash("aes", key)
     hmac_key = tagged_hash("hmac", key)
+    # constant-time compare (L6) - do not replace with == / !=
     if not consteq(mac, hmac.new(hmac_key, ct, digestmod="sha256").digest()):
         raise Exception("Invalid HMAC")
     b = BytesIO(ct)
--- a/src/helpers.py
+++ b/src/helpers.py
@@ -38,13 +38,17 @@ def tagged_hash(tag: str, data: bytes) -> bytes:
 
 def consteq(a: bytes, b: bytes) -> bool:
     """
-    Constant-time comparison for MACs, HMACs and other secret-derived
-    fixed-length values. Plain `==`/`!=` on bytes can stop as soon as a
-    differing byte is found, so its timing can leak how many leading
-    bytes matched. This walks every byte of both inputs unconditionally
-    and only combines the differences at the end, avoiding that timing
-    side channel. Used because the MicroPython build on this device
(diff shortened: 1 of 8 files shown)


Schnuartz:
@al-munazzim can you review?

al-munazzim (approved):
Reviewed the constant-time comparison change and the SecureChannel/PIN/storage-MAC call sites. The implementation keeps the fixed-length caveat explicit, uses the helper only for fixed-size authentication values, and CI is green.\n\nNo objections from my side.

src/helpers.py
```diff
@@ -36,6 +36,34 @@ def tagged_hash(tag: str, data: bytes) -> bytes:
...
+    Used because the MicroPython build on this device (custom C `hmac`
+    usermod) has no `hmac.compare_digest()`.
+
+    SECURITY: do not "simplify" the loop below back to `a == b` / `!=`,
+    and do not add an early `return False` inside the loop - either
+    change reintroduces the timing side channel this function exists
+    to remove. A test asserting equal() == equal() cannot catch that
+    regression, since both implementations return the same booleans.
+    """
+    if len(a) != len(b):
```
maggo83, line 59: Why don't we consider the length sensitive information?
If it is clear which of the inputs a/b is the "expected" and which is the "to_be_tested" we can have a function that is constant runtime for the length of "to-be-expected".
If to-be-expected is shorter: stop comparing at its length.
if to-be-expected-is longer: keep comparing to a dummy byte/wrapped expected (while forcing return to be "different")

test/tests/test_helpers.py
```diff
@@ -1,7 +1,43 @@
 from unittest import TestCase
-from helpers import conv_time
+from helpers import conv_time, consteq, aead_encrypt, aead_decrypt
 
 class HelpersTest(TestCase):
```
maggo83, line 4: Shouldn't we try to test that the runtime is constant somehow? Currently this is only asserted by review.
Maybe this could also be an automatic evaluation / review of the actual bytecode? [actual runtime measurements are going to be noisy I assume]

src/helpers.py
Schnuartz: I don’t think this would provide a security benefit here. All current consteq() call sites compare fixed, public-length authentication values: 32-byte SHA-256 HMACs or 14-byte SecureChannel MACs. If the received value has a different length, that length is already known to the party providing the input, so returning early on a length mismatch does not reveal any additional secret information.

test/tests/test_helpers.py
Schnuartz: I tried a lot for testing it. And I found out that it would be an unreasonable work to test that, even with the help of AI.

It works in theory, so we should merge it near time to solve the issue.

maggo83 (approved):

Schnuartz (approved):
This makes sense. 

Reviewed it. And it adds a security benefit.

PR #389, opened 2026-08-26
Proof of concept: Specter DIY on the Waveshare ESP32-P4 4.3-inch board

sandman21vs:
> [!WARNING]
> **This is a proof of concept. Do not use it with real funds.**
>
> Secure Boot and flash encryption are off, no eFuses are burned, and nothing
> here has been audited. The device PIN protects the interface, not the flash:
> anyone who desolders the chip reads the seed. The bootloader approval chain is
> not implemented, and `wipe()` deletes files rather than erasing the partition.
>
> It is opened to show that the port is viable and to share what the hardware
> bring-up turned up, not to be merged as-is.

Related to the ESP32-P4 work in `miketlk/specter-bootloader@port_esp32-p4`.

## Summary

Runs Specter DIY on the Waveshare ESP32-P4-WIFI6-Touch-LCD-4.3-C.

- `WAVESHARE_P4_43` MicroPython board, built from MicroPython `master` for the
  `PRE_REV3` chip variant, with a 4 MB app partition
- `p4board` component: ST7701 480x800 MIPI-DSI panel, LEDC backlight, GT911
  touch over raw I2C, and the companion ESP32-C6 held in reset at boot
- LVGL 9.3 rendering **directly into the panel framebuffer**, no copy on flush,
  reusing the bundle's pre-generated `lv_mpy.c` binding
- `secp256k1` and `uhashlib` user C modules translated from their Makefile-only
  builds to the CMake path the esp32 port uses
- `pyb` and `sdram` shims mapping the STM32-era API the app speaks onto
  `machine` and PSRAM, following the precedent of the simulator's 117-line shim
- microSD over SDMMC on GPIO 39-44
- OV5647 MIPI-CSI camera with `k_quirc` decoding, replacing the serial QR
  scanner this board does not have

The app changes are deliberately small: an `esp32` branch in `platform.py`
alongside the existing `simulator` one, a `simulator` predicate fix in
`config_default.py`, and three lines in `hosts/qr.py`.

## The camera replaces the scanner without touching QRHost

`QRHost` reads the scanner through `uart.any()` and `uart.read()`, expecting a
CR-terminated payload. Rather than rewriting the host, the shim hands it a
virtual UART fed by the camera. All of its protocol logic — animated QR, UR,
BBQr, part reassembly — keeps working, since it never cared where the bytes came
from.

The only edit in `hosts/qr.py` makes `init()` skip serial scanner probing when
the uart declares itself camera-backed. Without it the host burns timeouts
querying a scanner that is not there, and a QR in view during probing could be
mistaken for a scanner reply.

## Validation

Everything below was measured on the board, not inferred.

- [x] MicroPython REPL up, 31.6 MB PSRAM free, `PRE_REV3` variant confirmed in
      the banner
- [x] ST7701 panel at 480x800, test pattern rendered
- [x] GT911 touch: 819 points over 20 s at 50 Hz, spanning x 4-475, y 8-797
- [x] `sha256`, `sha512`, `ripemd160` against the standard `"abc"` digests
- [x] `hmac_sha512` against RFC 4231 test case 1
- [x] `pbkdf2_hmac` BIP39 seed
- [x] TRNG sanity: 50.0 % ones and all 256 byte values across 4096 bytes
- [x] **All seven official BIP84 test vectors** from `bitcoin/bips` reproduce,
      including `bc1qcr8te4kr609gcawutmrza0j4xv80jy8z306fyu`
- [x] OV5647 camera at 1280x960 RGB565, 45.5 fps
- [x] QR decoding at 11.3 scans/s, test code read in under 4 s
- [x] microSD: 2 GB card detected, mounted FAT, listed, written, read back,
      deleted, unmounted
- [x] Specter app boots and runs its asyncio loop on the panel

## What this branch depends on

- **MicroPython `master`.** `ESP32_GENERIC_P4` exists only there — it is absent
  from v1.25.0 and v1.26.0, so no released MicroPython can target this chip.
- **PR #358.** The GUI on `master` uses LVGL 6 API (`lv.style_copy`, `lv.btnm`,
  `lv.page`), removed in LVGL 8/9, while the maintained
  `lv_binding_micropython` targets LVGL 9.3. This branch merges the LVGL 9.3
  migration, so #358's commits appear here with their original hashes.
- **A fork of `secp256k1-embedded`**, carrying a one-line fix for an API that
  MicroPython `master` removed. Upstreaming that would drop the fork.

ESP-IDF stays at 5.5.x. MicroPython does not support the 6 series, and the
camera components turned out not to need it.

## `reports/`

Twelve findings from the bring-up, written up for the projects they belong to
rather than silently worked around:

| For | Finding |
| --- | --- |
| `odudex/Kern` | Hardware counter-evidence for the open boot-hang issue #135 on this exact board |
| `odudex/Kern` | Touch reset GPIO divergence, with a measurement showing the dual-address probe is what makes touch work here |
| `odudex/Kern` | `wave_43` declares no SD card, but the slot works — now verified with a card end to end |
| `miketlk/specter-bootloader` | `tools/idf.sh` fails silently on Linux: `dash` has no `$BASH_SOURCE` and the error goes to `/dev/null` |
| `lvgl/lv_binding_micropython` | The generated binding calls `mp_obj_int_to_bytes_impl()`, removed from MicroPython |
| `micropython/micropython` | User C module compile definitions do not reach the QSTR extraction pass on the CMake path |

Plus notes on the `cbor2` pin that will not install on Python 3.14, the
`k_quirc` licence being MIT despite GitHub reporting `NOASSERTION`, and a
cross-validation showing two independent implementations agree on every panel
timing for this board.

## Known gaps

- Flash protection status reports `unknown`. eFuses cannot be read from
  MicroPython, and reporting the build profile's intention as a runtime fact
  would be a lie in a security readout.
- `wipe()` removes files but does not erase the partition — that needs to happen
  in the bootloader, and it is marked not implemented rather than silently
  skipped.
- No smartcard reader on this board, so the Specter-Javacard keystore is stubbed
  to report no card and the app falls back to the SD and flash keystores.
- USB host is stubbed; the REPL arrives over the CH343 bridge and native USB is
  not wired.


sandman21vs:
Opened against the wrong repository by mistake — this was meant for my own fork. Closing. Sorry for the noise.

PR #392, opened 2026-08-28
feat: warn on disproportionately high transaction fees

schnuartz-ai:
## What this changes

When Specter parses a Bitcoin PSBT before signing, it now warns if the transaction fee is at least 10% of the amount being sent. The warning uses the existing transaction warning area on the confirmation screen.

The threshold is fixed at 10%, matching the BitBox fee-percentage threshold. There is no user setting for changing or disabling it.

## Fee-basis behavior

- The security calculation identifies outputs owned by an input wallet separately from the UI `change` classification.
- Owned outputs are excluded from the sent amount even for multi-wallet or collaborative transactions.
- If a transaction has no externally owned output, the verified total input amount is used as the fallback basis.
- The 10% comparison is integer-based and inclusive: `fee * 100 >= fee_basis * 10`.
- The displayed percentage is calculated from the same basis and includes the calculated percentage, for example: `Fee is 10.00% of the amount - unusually high!`.
- Existing transaction warnings are preserved, and the same warning is not added twice.
- Liquid transactions are unchanged because their fee is not available in the transaction metadata used here.

Manual test vectors are provided in a separate PR comment and are not part of this branch.

Closes #324

Timelock receovery (4)


schnuartz-ai:
## Manual test vectors for the high-fee warning

> PUBLIC DEMO SEED: never use this mnemonic for real funds. Anyone can reconstruct and control this wallet.

Tested commit: `9f182ffe56f725a4293290887471698356ae2cc9`

The vectors below were generated with `3rdIteration/psbt_faker`. They are regtest-only PSBTs with fictional inputs and must not be broadcast. Scan one complete QR image at a time.

### Demo wallet setup

Mnemonic:

`ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost ghost machine`

Standard SeedQR digits:

`078007800780078007800780078007800780078007801068`

SeedQR - white background

For a fresh regtest wallet, import the SeedQR above with an empty passphrase. The matching BIP84 descriptor is:

`wpkh([8c24a510/84h/1h/0h]tpubDC4DsqH5rqHqipMNqUbDFtQT3AkKkUrvLsN6miySvortU3s1LGaNVAb7wX2No2VsuxQV82T8s3HJLv3kdx1CPjsJ3onC1Zo5mWCQzRVaWVX/{0,1}/*)`

### Test 1: Below threshold

- Input total: `1,000,000 sats`
- External output: `910,000 sats`
- Fee: `90,000 sats`
- Fee ratio: `9.89%`
- Expected result: no high-fee warning

Test 1 - below 10 percent - white background

### Test 2: Exact threshold

- Input total: `1,100,000 sats`
- External output: `1,000,000 sats`
- Fee: `100,000 sats`
- Fee ratio: `10.00%`
- Expected result: high-fee warning

Test 2 - exactly 10 percent - white background

### Test 3: Above threshold

- Input total: `1,200,000 sats`
- External output: `1,000,000 sats`
- Fee: `200,000 sats`
- Fee ratio: `20.00%`
- Expected result: high-fee warning

Test 3 - above 10 percent - white background

The all-change fallback and multi-wallet basis were also exercised locally against the same PR commit: the fallback uses verified input total, while an owned output from an input wallet is excluded from the external send amount even when multiple wallets are present.

schnuartz-ai:
@al-munazzim can you review this?

al-munazzim:
Thanks @schnuartz-ai. I checked the current diff and ran the native suite locally in a fresh venv; it passes, but only 5 tests ran because the warning-specific native test file introduced in the first commit is removed again in the final diff.\n\nBefore merging, please add/restore automated coverage for the final behavior, especially:\n\n- inclusive 10% threshold (exactly 10% warns),\n- fee_basis fallback when there is no external recipient output,\n- owned outputs excluded from the fee basis even in multi-wallet/collaborative cases,\n- preserving existing warnings and avoiding duplicate high-fee warnings.\n\nThe manual QR vectors are useful for device validation, but I'd prefer not to rely on them as the only regression coverage for this transaction-safety warning.

maggo83:
Before checking code: shouldn't we have some exception from this warning for very small (test) transactions?

schnuartz-ai:
Pushed `1e2d3d1` addressing the review feedback. CI (native-tests, tests, Build Firmware) is green.

**Tests restored** — `test/tests_native/test_wallet_manager_warnings.py` now drives `add_warnings()` directly:
- fee below 10% → no warning; **exactly 10% → warning** (inclusive); above 10% → warning
- self-transfer (all outputs owned) → basis falls back to the **verified total of inputs**
- owned output excluded from the basis even with **multiple input wallets**
- existing warnings preserved; the high-fee warning is not added twice

**Display consistency** — `add_warnings()` now derives the fee basis itself (verified recipient outputs, or verified input total for self-transfers) and stores it in `meta["fee_basis"]`. `TransactionScreen` reads that same value, so the percentage it shows can no longer disagree with the warning. Output ownership is tracked separately from the UI `change` flag via a new `meta["outputs"][i]["owned"]`.

**Warning text** now states the basis: `Fee is X% of the send amount - unusually high!` vs `Fee is X% of total inputs (self-transfer) - unusually high!`

**Unmissable** — a high fee must be acknowledged on its own `Prompt` in `confirm_transaction()` before the final review screen; it can't be scrolled past. It still also appears in the on-screen warnings list.

Threshold stays fixed at 10% with no user setting; re: the small-transaction question, no exemption was added (out of scope for this PR).

Schnuartz:
> Before checking code: shouldn't we have some exception from this warning for very small (test) transactions?

No, don't think so. BitBox doesn't do this as well. And it's just a warning you can click away if you are making small transactions.

schnuartz-ai:
Follow-up `3337c88`: strengthened the acknowledgement test coverage.

The earlier test only asserted `meta["fee_warning"]` was set, so it would have stayed green if the `confirm_fee_warning()` call were dropped from `confirm_transaction()`. Added two tests that drive `confirm_transaction()` with the sub-steps stubbed:

- fee **not** acknowledged → `confirm_transaction()` returns `None` and the final confirmation screen is never reached
- fee acknowledged → final confirmation is reached and signing options are returned

Confirmed locally that removing the gate makes the first test fail. CI green (native-tests, tests, Build Firmware).

Still recommended before merge: a hardware smoke test on a Specter, since the new UI (separate warning Prompt + `of send amount` / `of total inputs` fee text) isn't covered by the manual QR vectors, which were captured against `9f182ff`:
scan a PSBT with fee ≥10% → dedicated warning Prompt appears → Cancel aborts → re-scan → Proceed → normal transaction confirmation → signing works.

Schnuartz (approved):
Checked it and approve. this change makes sense and I reviewed it. 

I did not tested on RL Hardware yet. But as it is a tiny change and we have simulator changes it is not necessary.

If someone wants to do it. It is very easy as there is a test in the comments.

src/apps/wallets/manager.py
maggo83, line 1: HighFee Warning is not carried over to liquid. No reason why this should be bitcoin specific

src/gui/screens/transaction.py
```diff
@@ -75,10 +72,16 @@ def __init__(self, title, meta):
...
         if fee:
-            if send_amount > 0:
-                fee_percent = fee * 100 / send_amount
-                fee_txt = "%d satoshi (%.2f%%)" % (fee, fee_percent)
-            # back to wallet
+            # use the same verified basis the high-fee warning is computed
+            # from, so the displayed percentage can never disagree with it
+            fee_basis = meta.get("fee_basis", 0)
+            if fee_basis > 0:
+                fee_percent = fee * 100 / fee_basis
```
maggo83: Nit: why not precompute fee_percent already in wallet and then re-use it here? That woudl guarantee absolute consistency

maggo83 (commented):
We should have a conscious decision about why this should not be carried over to liquid.

src/apps/wallets/manager.py
schnuartz-ai: Fixed in 349125a. `LWalletManager.preprocess_psbt()` overrode the base method wholesale and never called `add_warnings()`, so the warning was structurally unreachable on Liquid (not a deliberate choice).

Now:
- Liquid `preprocess_psbt()` calls `add_warnings()`.
- `meta["fee"]` is taken from the explicit L-BTC fee output (empty scriptPubKey), which is flagged `fee_output`.
- A `fee_basis()` override keeps the basis restricted to the fee (L-BTC) asset so multi-asset transfers aren't mixed in, and skips the fee output itself.
- `confirm_transaction()` (with the acknowledgement gate) is inherited unchanged, so the mandatory Prompt now fires for Liquid too.

Still worth a manual Liquid smoke test (PSET with fee >=10% -> warning Prompt -> Cancel aborts -> Proceed -> normal confirmation).

src/gui/screens/transaction.py
schnuartz-ai: Done in 349125a. `add_warnings()` now computes the percentage once and stores it as `meta["fee_percent"]` (raw float, whenever `fee_basis > 0`, independent of the threshold check). `TransactionScreen` reads that value instead of recomputing `fee * 100 / fee_basis`, so the number in the warning text and the number on the confirmation screen can no longer diverge. Storing the float (not a formatted string) keeps display formatting in the screen.

schnuartz-ai:
@maggo83 both review points are addressed in 349125a:

- **Liquid parity** — the warning wasn't Bitcoin-specific by design; `LWalletManager.preprocess_psbt()` overrode the base method wholesale and never called `add_warnings()`. It now does, derives `meta["fee"]` from the explicit L-BTC fee output, and a `fee_basis()` override keeps the basis on the fee asset. The acknowledgement Prompt fires for Liquid too.
- **fee_percent consistency** — computed once in `add_warnings()` as `meta["fee_percent"]`, read verbatim by `TransactionScreen`, so warning text and on-screen percentage can't disagree.

Native tests extended for both. A manual Liquid smoke test is still recommended before merge.

src/apps/wallets/liquid/manager.py
```diff
@@ -569,13 +569,45 @@ def preprocess_psbt(self, stream, fout):
...
             out.write_to(fout, skip_separator=True, version=psbtv.version)
             # write rangeproofs and surjection proofs
             # separator
             fout.write(b"\x00")
 
+        self.add_warnings(meta)
         return wallets, meta
 
+    def fee_basis(self, meta):
+        """Same verified basis as the base class, but restricted to the
```
maggo83: a) much code duplication from fee_basis in btc only case: I'd think about how to unify the code basis more.
b) since only outputs with the same asset used for the fee counts toward the send_amount/fee_basis: in case one uses one (or several) assets (some liquid tokens) to send and another asset (say l-btc) to pay the fees, this warning will easily give wrong warnings / possible false alarms, as we do not (cannot?) compare the "real world" value of the actual transferred assets to the fee being sent. Maybe we should have a separate warning for this case (=the fee ouput is the only output that is denominated in the fee asset), that we cannot assess if the fee is "appropriate" and that the user should double check?

Schnuartz: I understand your point. However, I am currently communicating that Liquid support is not maintained.
Therefore not implementing Liquid properly at some points in code is ok. 

It would needs to be cleaned up in the future anyway. And we shouldn't fokus on it as long as blockstream wallet is not supporting Specter DIY.

schnuartz-ai: Thanks, I agree with both points.
I unified the common fee-basis calculation and kept only the asset-specific filtering in the Liquid manager.
For Liquid, the normal percentage warning is shown only when the fee asset is comparable to the transferred assets. If the transaction transfers a different asset from the one used to pay the fee, Specter now shows a separate warning stating that the fee cannot be assessed automatically and should be checked manually.
This way, both Bitcoin and Liquid transactions provide a warning without comparing unrelated asset amounts.

maggo83 (approved):

PR #396, opened 2026-08-29
Fix incomplete secure wipe of QSPI flash

schnuartz-ai:
## Problem

`platform.wipe()` deleted the QSPI filesystem but physically overwrote only the internal-flash range. Most of the 16 MiB QSPI user-data region could therefore remain recoverable from the chip.

## Resolution

- Replace the hardcoded range with the verified internal-flash and full-QSPI block map.
- Overwrite both regions in bounded chunks without loading the QSPI region into RAM.
- Validate flash geometry before destructive work and explicitly flush write-behind caches.
- Track write, sync, unmount, and wipe failures; never treat a partial wipe as successful.
- Build both `USE_DBOOT=0` and `USE_DBOOT=1` firmware configurations in CI.

## Note: this replaces #386

#386 was closed by accident, and the head branch was deleted in the same action. GitHub then refuses to reopen it: both `gh pr reopen` and the REST endpoint fail (`422 Validation Failed`), and the PR page offers no reopen control at all. Restoring the head branch to the exact commit it was closed at (`b93cdca`, via the PR page's own "Restore branch" button) did not lift the restriction. This PR therefore continues the same branch, `upstream/pr12-qspi-secure-wipe`, with no changes made for the move.

Please read #386 for the full review discussion — it is still worth reading and I have not copied it here.

## Review status carried over from #386

- @al-munazzim approved on 2026-08-26 after checking out the branch and running `python test/run_native_tests.py` (31 tests passing, CI green).
- @maggo83 raised five points on `src/platform.py`: comment length and fragile references into submodule paths, constants duplicating information the project already has, placement of the lower-level helpers relative to the existing architecture, wipe behaviour when `ioctl` reports unexpected sizes, and whether the `ok` flag should start pessimistic.

All five were answered in #386, and two led to code changes that are included here:

- `b93cdca` moves the block-map derivation into `docs/flash-block-map.md`, leaving a short block map plus a pointer in the code. The submodule references now live in the doc with a note to re-verify them on a submodule bump.
- `cd1012a` removes the block-size and QSPI-end constants; `wipe()` reads both back from the block device via `ioctl()`. The split point between `/flash` and `/qspi` remains, since no `ioctl` exposes it.

On the architectural placement I opened diybitcoinhardware/f469-disco#44 to get this exposed as a proper board-level API. Until that exists there is nothing to import, so the split point stays here for now.

The two remaining points (`ioctl`-controlling attacker, pessimistic `ok` flag) I argued against in #386 rather than changing; @maggo83, those are still open for you to push back on.

## Tests

The port includes native regression coverage for block boundaries, chunking, failures, geometry checks, sync ordering, and simulator behavior. CI also builds both firmware configurations.

This change was originally developed and tested in [Schnuartz/specter-diy#12](https://github.com/Schnuartz/specter-diy/pull/12). The original PR contains the full development history, test results, review discussion, and implementation details. The results below refer to the original source revision. Fresh CI on this upstream PR validates the cleanly ported revision against the current upstream master.

- `python3 test/run_native_tests.py`: 31/31 passed
- Both firmware configurations (`USE_DBOOT=0` and `USE_DBOOT=1`) were built/tested on the source branch

The original #12 source revision was not tested on real hardware for completion of the full 16 MiB QSPI wipe. No hardware validation is claimed for this clean upstream revision. The pinned MicroPython drivers also do not expose every possible underlying erase/program failure to Python; this limitation is documented in the implementation.

### DIFF
--- a/src/platform.py
+++ b/src/platform.py
@@ -2,7 +2,6 @@
 import sys
 import os
 import pyb
-import gc
 
 simulator = (sys.platform in ["linux", "darwin"])
 
@@ -392,10 +391,13 @@ def reboot():
 # remapping or translation layer sits between the block numbers above and
 # the physical sectors (see spi_bdev_writeblocks()/mp_spiflash_cached_write()
 # in ports/stm32/spibdev.c and drivers/memory/spiflash.c), so overwriting
-# every block in a range does physically destroy the data that was there.
-# Required sector-erase-before-write is handled transparently by that
-# driver. Sources checked: diybitcoinhardware/f469-disco (submodule pin
-# db3ce3e), diybitcoinhardware/micropython (submodule pin 6bdf1b6),
+# every block in a range overwrites all addressable QSPI sectors, which
+# prevents recovery through normal digital reads of the flash contents
+# (this says nothing about charge-remanence/decapping-level analog
+# recovery, which is out of scope for this threat model). Required
+# sector-erase-before-write is handled transparently by that driver.
+# Sources checked: diybitcoinhardware/f469-disco (submodule pin db3ce3e),
+# diybitcoinhardware/micropython (submodule pin 6bdf1b6),
 # ports/stm32/{storage.h,storage.c,flashbdev.c,spibdev.c},
 # ports/stm32/boards/STM32F469DISC/{mpconfigboard.h,mpconfigboard.mk},
 # ports/stm32/boards/stm32f469xi_dboot.ld, drivers/memory/spiflash.c.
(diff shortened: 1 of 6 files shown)


al-munazzim (approved):
Re-reviewed this continuation of #386. I checked the current PR state, CI is green, and I ran the native suite locally in a fresh venv: `python test/run_native_tests.py` => 35 tests passing. The latest changes address the prior review concerns around deriving geometry, moving the block-map derivation into docs, and sync ordering. I'm still okay with this direction; the hardware-validation caveat in the PR description remains the main limitation.

Schnuartz (approved):

src/platform.py
```diff
@@ -357,32 +356,224 @@ def reboot():
...
 
+# Verified block map of pyb.Flash() on the STM32F469 Discovery board
+# (make disco USE_DBOOT=1). Absolute block numbers:
+#   block 0            emulated MBR, not backed by real flash
+#   blocks 1   - 255    unmapped (bootloader/reserved sectors), never
+#                       reachable through this API
+#   blocks 256 - 447    internal flash, the "/flash" filesystem (96 KiB)
+#   blocks 448 - 33215  external QSPI flash, the "/qspi" filesystem (16 MiB)
+# Full derivation, source files checked and submodule pins:
+# see docs/flash-block-map.md
```
maggo83, line 367: still duplicated compared to file

src/platform.py
```diff
@@ -357,32 +356,224 @@ def reboot():
...
+# The boundary between the two filesystems is the only part of the map
+# below that no ioctl() exposes: it comes from the linker script and
+# storage.h inside the pinned submodule, so it has to be stated here.
+# Block size and the end of the QSPI region are *not* restated - wipe()
+# reads both from the block device at run time (see
+# diybitcoinhardware/f469-disco#44 for making the boundary queryable too).
+INTERNAL_FLASH_START_BLOCK = 0x100                             # FLASH_PART1_START_BLOCK
+INTERNAL_FLASH_END_BLOCK = INTERNAL_FLASH_START_BLOCK + 192     # + FLASH_MEM_SEG1_NUM_BLOCKS
```
maggo83, line 375: 192 = magic number

maggo83 (approved):

PR #401, opened 2026-08-30
fix: switch host to 9600 baud after GM65 factory reset

### LINKED ISSUE #355: Problems with PR #299 | GM65 Scanner Module, SW 137, SW 134
When pressing: 

_Device settings -> Communication -> QR Scanner -> "Reset QR Module"_

With the following Scanner Module Firmwares (bzw. the very old GM65 Scanner Modules)
GM65 SW 137 and GM65 SW 134

Then it's showing:

But it's still resetting something causing it to not scanning it until you restart the device which fixes the scanner again. 

I was reproduce that with the reset QR-Code from the GM65 Manual at 1.7 as well:
https://github.com/3rdIteration/SerialBarcodeReaderTools/blob/master/Manuals/GM65/GM65-S-UserManual.pdf
image

After the first bug the scanner also said something like "Failed to configure Scanner" when you want to change the scanner settings until restarting the device which makes everything work normal again.

schnuartz-ai:
## Problem

Triggering "Reset QR Module" on a GM65 left the scanner unable to deliver
data: it lit up and beeped on a successful decode, but nothing reached the
device, and only a power cycle helped. That is the symptom reported in #355.

Four separate defects were involved. Only the first is about baud rates; the
rest are in the reset path itself and predate this PR.

**1. The host-side baud switch was gated on the reset ACK.**
A GM65 factory reset restores the module to its defaults, which per Form 2-1
of the GM65-S manual are standard TTL-232 at 9600 baud. The reset is
fire-and-forget: once the command is on the wire the scanner is at 9600
whether or not the ACK survives the baud change. Gating the host-side switch
on the ACK skipped it in exactly the case #355 describes, and
`configure_gm65()` does no baud probing, so nothing recovered.

**2. `DELAY_AFTER_FACTORY_RESET` (200 ms) does not cover the module reboot.**
A factory reset reboots the GM65. The reconfiguration started writing while
the module was still coming up, so it silently did not land. This was the
dominant failure on hardware.

**3. The baud guard skipped `clean_uart()`.**
`if res and self.baudrate != BAUD_RATE_9600` meant `_set_baud()` — and with
it the UART drain — was skipped whenever the host was already at 9600, which
is precisely the boot-reset path since `_update_scanner_model()` probes 9600
first. Boot noise the module emits past the seven-byte ACK then got read as
the first reply of the reconfiguration.

**4. A failed reconfiguration left a wiped, unconfigured scanner behind.**
`init()` falls back to PIN trigger mode when `configure()` fails. The reset
path did not, and `Host.enable()` only calls `init()` once per session, so
nothing retried — hence "only a reboot helps".

Two further problems found while testing this on hardware:

**5. The dialog said "Factory reset" / "Resetting scanner to defaults..."**
on a device whose main job is holding a seed, with nothing in the dialog to
say only the scanner was affected.

**6. The scanner info line read the host's `raw_fix_applied` setting**, which
survives a reset that wiped RAW mode on the scanner, so it reported
`CompactQR fix: Applied` for a scanner that had just lost it.

## Fix

* switch the host to 9600 unconditionally after sending the reset, and decide
  success by whether the scanner answers there afterwards
* poll for the module to come back (`SCANNER_REBOOT_TIMEOUT_MS`) instead of a
  fixed 200 ms
* always go through `_set_baud()`, so the UART is drained
* share `init()`'s PIN-trigger fallback with the reset path
* report the RAW state last verified on the scanner, not the host flag
* name the scanner in the reset dialog and state that keys and device
  settings are untouched
* restore the scanner preferences (sound, aim light, flashlight) to their
  defaults, so a "factory reset" means what it says. `enabled` is left alone:
  it controls whether the QR button appears in the main menu rather than how
  the scanner behaves

## Testing

`test/tests_native/test_qr_factory_reset.py` drives a fake GM65 that models
the two things a naive UART mock hides: it only answers when the host baud
matches its own, and a reset makes it unresponsive for a configurable window
before it comes back on the Form 2-1 defaults, emitting boot noise that is
not a protocol reply. A virtual clock replaces `time.sleep_ms`, so the reboot
window is covered without the suite sleeping. Every defect above fails
against the commit that precedes its fix.

    python3 test/run_native_tests.py                                  # 89 passed
    ../f469-disco/micropython/ports/unix/micropython run_tests.py     # 19 passed

Hardware, STM32F469I-DISCO, firmware built from f53f3d8:

* older-generation GM65 — works
* newer-generation GM65, SW 105 (`0x69`, needs the RAW fix) — works
* M3Y — works

Scanning after the reset works without a power cycle, CompactSeedQR included.
The SW 105 scanner is the one that reproduced the original failure before
these commits.

### Correction to an earlier commit message

The message of commit `59a3c6a` claims hardware verification of the
`== SUCCESS` check. That claim is not supported: the hardware test it refers
to ran several hours before that commit existed, against `0cacf76`. The
commit is kept for history and its effect is undone here; please read its
verification claim as void.

### If a GM65 ends up in this state

It is recoverable without firmware. Display the "Reset" setup barcode from
section 1.7 of the GM65-S User Manual and let the scanner read it, then power
cycle the device. Setup codes are enabled by default (section 1.5), so a
scanner that still decodes will accept it. Section 2.1 additionally has
"Series Output" and "9600bps(Default)". Verified on hardware. That barcode
encodes `RD9FF55` — the same zone bit and value `FACTORY_RESET_CMD` writes,
which confirms the two are the same operation.

Fixes #355.


schnuartz-ai:
#401 - PR - switch host to 9600 baud after GM65 factory reset
Analyse von https://github.com/cryptoadvance/specter-diy/pull/401
https://github.com/cryptoadvance/specter-diy/pull/401

## Zusammenfassung

Der PR fügt im GM65-Zweig von `_send_factory_reset` (`src/hosts/qr.py:601-607`) einen Wechsel der Host-UART auf 9600 Baud ein, sobald der Factory-Reset erfolgreich quittiert wurde. Zuvor wurde die Baudrate nach einem GM65-Reset gar nicht angepasst — der Scanner fällt auf 9600 zurück, der Host bleibt auf 115200, und jede folgende Kommunikation (`configure_gm65`) schlägt fehl. Das ist exakt die in Issue #355 beschriebene Symptomatik (Reset "macht etwas", Scanner scannt nicht mehr, "Failed to configure Scanner!", erst ein Neustart heilt). Die eigentliche Root-Cause-Analyse dazu liefert der ausführliche Kommentar zu Issue #355, auf den ich mich stütze.

## Sicherheit

Keine sicherheitsrelevante Auswirkung. Die Änderung betrifft ausschließlich das UART-Baudraten-Management auf Host-Seite. Keine kryptografischen Pfade, keine Vertrauensgrenze, keine neue externe Eingabe. `FACTORY_RESET_CMD` und alle Konstanten bleiben unverändert.

## Korrektheit

- **9600 als GM65-Default ist belegt:** `src/hosts/qr.py:44-45` ("After basic scanner configuration (9600 bps) uart is set to 115200 bps …"). `configure_gm65()` liest/schreibt zunächst bei der aktuell eingestellten Baudrate und schaltet erst ganz am Ende auf 115200 (`src/hosts/qr.py:480`). Die Baud-Einstellung (`BAUD_RATE_ADDR`, Zeile 477) erfolgt *nach* `save_settings_on_scanner()` (Zeile 451-455) und ist nicht Teil des gespeicherten Batches (Zeile 395-402). Damit enthält das EEPROM die 9600, und `FACTORY_RESET_CMD` (Datenbyte `0x55` = "Restore to user setting", laut Kommentar Zeile 35) stellt verlässlich 9600 wieder her. Die Prämisse des PR ist also korrekt.
- **`_set_baud` ändert die Host-Baud real:** `src/hosts/qr.py:483-487` deinit→`self.baudrate=baudrate`→`init(baudrate=…)`→`clean_uart()`. `self.baudrate` wird ausschließlich hier geschrieben und bleibt synchron mit dem UART — keine Staleness. Der Guard `if res and self.baudrate != BAUD_RATE_9600` ist eine korrekte Optimierung (kein unnötiger Reinit bei bereits passender Baud).
- **`self.baudrate`-Pflege:** zuverlässig, da jeder Baudwechsel über `_set_baud` läuft (Suche über die Datei bestätigt das).

## Architektur / Konsistenz mit dem M3Y-Zweig

Der M3Y-Zweig (`src/hosts/qr.py:586-599`) geht strategisch *anders* vor: er stellt die Baud *vor* dem Reset auf 9600 (`M3Y_BAUDRATE_SET`), sendet den Reset und stellt danach `prev_baudrate` wieder her. Der GM65-Zweig belässt den Host nach dem Reset bei 9600. Beides ist für den jeweiligen Scanner korrekt:
- M3Y lässt sich per Text-Kommando auf jede Baud stellen, daher kann nach dem Reset wieder auf 115200 gewechselt werden.
- GM65 wird im Anschluss über `_apply_post_reset_configuration` → `configure()` → `configure_gm65()` neu konfiguriert; `configure_gm65` erwartet, bei 9600 zu beginnen, und schaltet am Ende auf 115200. Den Host bei 9600 zu belassen, ist also genau der richtige Zustand für den Folge-Schritt.

Die Zweige sind also nicht identisch, aber jeweils passend. Kein Konsistenzproblem.

## Fehlerpfade & Edge Cases

- **Reset schlägt fehl (`res` False/None):** `if res and …` short-circuitiert, `_set_baud` wird nicht gerufen, Baud bleibt unverändert, `return res` → `False` → Anzeige "Failed to factory reset scanner!" (`src/hosts/qr.py:695`). Korrekt: kein eingeführter Desync.
- **Potenzielle Lücke — Reset wirkt, aber ACK fehlt:** `query()` (`src/hosts/qr.py:212-223`) gibt `None` zurück, wenn innerhalb des Timeouts keine Daten kommen (Zeile 216-217). `bool(None)` = `False`. Wenn also der Scanner den Reset ausführt (→ 9600) *ohne* nutzbaren ACK zu senden — z.B. weil er vor der Antwort resetet, oder weil die Antwort bei noch auf 115200 stehendem Host garbled ankommt — dann ist `res` False, der Host bleibt auf 115200, der Scanner ist auf 9600: der Desync bleibt bestehen und der PR greift nicht. Für die im Issue genannten alten Firmwares (SW 137, SW 134) ist unklar, ob sie den Reset zuverlässig quittieren. Der PR ist in diesem Szenario nur bedingt wirksam. Abhilfe wäre ein fallback-Probieren auf 9600 auch bei `res == False` (analog zu `_update_scanner_model`, `src/hosts/qr.py:497-515`). Das ist kein Regressor — der Status quo ist hier *auch* kaputt —, aber der PR löst ggf. nicht das komplette Issue. **Empfehlung:** dies als bekannte Limitation im PR kommentieren oder den Fallback ergänzen.
- **Race Conditions:** Im async-Pfad (`_factory_reset_scanner`, `src/hosts/qr.py:626-638`) findet während `_send_factory_reset` kein `await` statt; `await asyncio.sleep_ms(DELAY_AFTER_FACTORY_RESET)` (Zeile 630) läuft erst *nach* dem Baudwechsel. Kein Zeitfenster, in dem eine andere Coroutine auf dem falsch konfigurierten UART reden könnte. Keine Race.
- **Exception Safety:** Wenn `_set_baud` nach `uart.deinit()` bei `uart.init()` wirft, bleibt der UART deinitiiert, während `self.baudrate` schon auf 9600 steht. Das ist ein bestehendes Risiko von `_set_baud` (gleichartig bei `src/hosts/qr.py:480`), kein Regressor durch diesen PR.

## Tests

Für den QR-Scanner-Host gibt es im Repo keine Tests (keine Treffer unter `tests/` für `_send_factory_reset`/`configure_gm65`/`GM65`/`_set_baud`). Der PR fügt keine hinzu. Wegen `pyb.UART`-Abhängigkeit ist echtes Unit-Testing aufwändig, aber mit einem UART-Mock liesse sich der Baud-Switch nach erfolgreichem Reset sowie das Ausbleiben bei `res == False` gut prüfen. **Vorschlag:** Mock-basierten Test für `_send_factory_reset` (GM65-Pfad) nachreichen; nicht blockierend, da konsistent mit dem bestehenden (testfreien) Stand des Hosts.

## Merge-Empfehlung

**Merge empfohlen.** Der Patch ist minimal, korrekt kommentiert, adressiert die identifizierte Root Cause für den Regelfall (Reset wird quittiert) und führt keine Regressionen oder Sicherheitsprobleme ein. Er ist konsistent mit dem 9600-Modell des GM65 im bestehenden Code. Vor einem finalen Merge sollte die Maintainerin/den Maintainer die "Reset wirkt, aber ACK fehlt"-Lücke (siehe Fehlerpfade) bewerten — ggf. durch einen 9600-Fallback bei `res == False` —, um Issue #355 auch für die alten Firmwares SW 137/134 vollständig zu schließen. Fehlende Tests sind im Kontext akzeptabel, aber als Follow-up wünschenswert.

schnuartz-ai:
#401 - PR - switch host to 9600 baud after GM65 factory reset
Analysis of https://github.com/cryptoadvance/specter-diy/pull/401
https://github.com/cryptoadvance/specter-diy/pull/401

## Summary

The PR adds a host UART switch to 9600 baud in the GM65 branch of `_send_factory_reset` (`src/hosts/qr.py:601-607`) once the factory reset is successfully acknowledged. Previously, the baud rate was not adjusted after a GM65 reset — the scanner falls back to 9600, the host stays at 115200, and all subsequent communication (`configure_gm65`) fails. This is exactly the symptom described in issue #355 (reset "does something", scanner no longer scans, "Failed to configure Scanner!", only a reboot fixes it).

## Security

No security-relevant impact. The change affects only host-side UART baud-rate management. No cryptographic paths, no trust boundary, no new external input. `FACTORY_RESET_CMD` and all constants remain unchanged.

## Correctness

- **9600 as GM65 default is documented:** `src/hosts/qr.py:44-45` ("After basic scanner configuration (9600 bps) uart is set to 115200 bps …"). `configure_gm65()` first reads/writes at the currently set baud rate and only switches to 115200 at the very end (`src/hosts/qr.py:480`). The baud setting (`BAUD_RATE_ADDR`, line 477) happens *after* `save_settings_on_scanner()` (lines 451-455) and is not part of the saved batch (lines 395-402). Thus the EEPROM contains 9600, and `FACTORY_RESET_CMD` (data byte `0x55` = "Restore to user setting", per comment at line 35) reliably restores 9600. The PR's premise is correct.
- **`_set_baud` actually changes the host baud:** `src/hosts/qr.py:483-487` does deinit → `self.baudrate=baudrate` → `init(baudrate=…)` → `clean_uart()`. `self.baudrate` is only written here and stays in sync with the UART — no staleness. The guard `if res and self.baudrate != BAUD_RATE_9600` is a correct optimization (no unnecessary reinit when baud already matches).
- **`self.baudrate` maintenance:** reliable, since every baud change goes through `_set_baud` (confirmed by searching the file).

## Architecture / Consistency with the M3Y Branch

The M3Y branch (`src/hosts/qr.py:586-599`) takes a strategically *different* approach: it sets the baud to 9600 *before* the reset (`M3Y_BAUDRATE_SET`), sends the reset, then restores `prev_baudrate`. The GM65 branch leaves the host at 9600 after the reset. Both are correct for their respective scanners:
- M3Y can be set to any baud via text command, so after the reset it can switch back to 115200.
- GM65 is subsequently reconfigured via `_apply_post_reset_configuration` → `configure()` → `configure_gm65()`; `configure_gm65` expects to start at 9600 and switches to 115200 at the end. Leaving the host at 9600 is exactly the right state for the next step.

The branches are not identical, but each is appropriate. No consistency problem.

## Error Paths & Edge Cases

- **Reset fails (`res` False/None):** `if res and …` short-circuits, `_set_baud` is not called, baud stays unchanged, `return res` → `False` → display "Failed to factory reset scanner!" (`src/hosts/qr.py:695`). Correct: no introduced desync.
- **Potential gap — reset takes effect but ACK is missing:** `query()` (`src/hosts/qr.py:212-223`) returns `None` if no data arrives within the timeout (lines 216-217). `bool(None)` = `False`. If the scanner executes the reset (→ 9600) *without* sending a usable ACK — e.g. because it resets before responding, or because the response arrives garbled while the host is still at 115200 — then `res` is False, the host stays at 115200, the scanner is at 9600: the desync persists and the PR doesn't take effect. For the old firmwares mentioned in the issue (SW 137, SW 134) it's unclear whether they reliably acknowledge the reset. The PR is only partially effective in this scenario. A remedy would be a fallback probe at 9600 even when `res == False` (analogous to `_update_scanner_model`, `src/hosts/qr.py:497-515`). This is not a regression — the status quo is *also* broken here — but the PR may not fully resolve the issue. **Recommendation:** note this as a known limitation in the PR or add the fallback.
- **Race conditions:** In the async path (`_factory_reset_scanner`, `src/hosts/qr.py:626-638`) no `await` happens during `_send_factory_reset`; `await asyncio.sleep_ms(DELAY_AFTER_FACTORY_RESET)` (line 630) runs only *after* the baud switch. No window where another coroutine could talk on the misconfigured UART. No race.
- **Exception safety:** If `_set_baud` throws after `uart.deinit()` at `uart.init()`, the UART remains deinitialized while `self.baudrate` is already set to 9600. This is a pre-existing risk of `_set_baud` (same at `src/hosts/qr.py:480`), not a regression introduced by this PR.

## Tests

There are no tests for the QR scanner host in the repo (no hits under `tests/` for `_send_factory_reset`/`configure_gm65`/`GM65`/`_set_baud`). The PR adds none. Due to the `pyb.UART` dependency, real unit testing is involved, but with a UART mock the baud switch after a successful reset and its absence when `res == False` could be verified. **Suggestion:** add a mock-based test for `_send_factory_reset` (GM65 path) as a follow-up; non-blocking, consistent with the existing (test-free) state of the host.

## Merge Recommendation

**Merge recommended.** The patch is minimal, correctly commented, addresses the identified root cause for the common case (reset is acknowledged), and introduces no regressions or security problems. It is consistent with the 9600 model of the GM65 in the existing code. Before a final merge, the maintainer should evaluate the "reset takes effect but ACK is missing" gap (see Error Paths) — potentially via a 9600 fallback when `res == False` — to fully close issue #355 also for the old SW 137/134 firmwares. Missing tests are acceptable in context but desirable as a follow-up.

Schnuartz:
Had huge problems testing that. Not ready to merge yet

Schnuartz:
Now tested it and worked. Message was sucessfull and GM65 Scanner worked normally afterwards.

But later on had problems with the QR-Scanner. Would not merge yet, until we know what is exactly going on.
IMG_20260902_174250

schnuartz-ai:
### Hardware verification

Tested on an STM32F469I-DISCO with firmware built from `f53f3d8`, across
three scanners:

| Scanner | Result |
|---|---|
| GM65, older generation | works |
| GM65, newer generation, SW 105 (`0x69` — needs the RAW fix) | works |
| M3Y | works |

The SW 105 unit is the one that reproduced the original failure before these
commits: the reset reported success, after which the scanner lit up and
beeped on a decode but no data reached the device, and only a power cycle
recovered it.

### Scope of that test, precisely

`f53f3d8` is this branch immediately before it was merged with master.
`src/hosts/qr.py` is byte-identical at `f53f3d8` and at the current head
`1e637f1`, so the code this PR is about was exercised exactly as it stands.

What the tested binary did *not* contain is everything the merge brought in
from master afterwards — `src/rng.py`, `src/apps/wallets/*`,
`src/gui/screens/transaction.py`, and the `f469-disco` bump from `db3ce3e`
to `9dd8515` (embit v0.8.2). Those are master's own already-merged changes,
not this PR's, but flagging it so nobody reads the table above as covering
the merged tree.

### A correction to an earlier commit message in this PR

The message of `59a3c6a` claims hardware verification of the `== SUCCESS`
check. That claim is not supported: the test it refers to ran hours before
that commit existed, against `0cacf76`. The commit is kept for history and
its effect is undone later in the branch — please read its verification
claim as void. Noted in the description as well.

### Recovery, if anyone hits this while reviewing

A GM65 left in the broken state is recoverable without firmware: display the
"Reset" setup barcode from section 1.7 of the GM65-S User Manual and let the
scanner read it, then power cycle. Setup codes are on by default (section
1.5), so a scanner that still decodes will accept it. That barcode encodes
`RD9FF55` — the same zone bit and value `FACTORY_RESET_CMD` writes, which is
independent confirmation that the two are the same operation.

Schnuartz (approved):

Issue #398, opened 2026-08-30
RAM shortage issue


Schnuartz:
I got some responses from Specter Users that we have to less RAM. 

Also had the Problem when Hardware testing this PR:
https://github.com/cryptoadvance/specter-diy/pull/369
<img width="4344" height="5792" alt="Image" src="https://github.com/user-attachments/assets/1432cc05-cbee-4f3e-a50a-25e856568bb6" />

But also have heard by signing bigger transactions or importing bigger Multi Signatur Wallets

PR #408, opened 2026-09-01
fix: harden secure storage deletion and wipe handling

schnuartz-ai:
## Summary

Harden secure storage deletion, SD-card erase/format, and device-wipe error handling. The mnemonic storage path is deliberately non-transactional across power loss: there is no `.tmp`/`.old` recovery protocol.

## Key changes

- Save encrypted mnemonics directly with strict synchronization.
- On confirmed replacement, best-effort logically overwrite and remove the existing file first, then write the new encrypted file.
- Delete mnemonics and other sensitive files by checking write sizes and reported sync errors before unlinking.
- Treat secure deletion as best-effort logical overwrite, not forensic erasure: the current MicroPython VFS/block-device layer may not surface every physical I/O failure, and wear-levelling may retain historical copies.
- Stream SD-card key detection with O(1) extra RAM and fall back to the known flash state when SD probing fails.
- Erase logically reachable SD-card blocks before recreating the filesystem, with validated geometry and explicit mount, unmount, write, and format error handling.
- Keep wallet and PSBT cleanup on the overwrite-before-unlink path.
- Wipe the documented internal-flash range, flush before reboot, and refuse to report a successful wipe when a reported write or flush fails.

A power interruption during save, replacement, deletion, SD formatting, or wipe may lose local data. Users must maintain an independent recovery backup of their recovery phrase.

## QSPI and key boundaries

The device wipe destroys the internal-flash keystore material, including the device secret and encryption secret. QSPI filesystem entries are logically removed and the filesystem metadata blocks are overwritten, but the QSPI data area is not physically overwritten.

Some QSPI data, such as host/global settings, uses keys derived from the device secret. Wallet descriptors and metadata are different: they use the seed-derived `idkey` at `m/0x1D'`. Destroying the device secret therefore does not by itself cryptographically erase those wallet files. Without the original seed they remain undecryptable, but this is not a blanket physical or forensic erasure of QSPI.

## Verification

- GitHub Actions for `1039d5b`: Build Firmware, native-tests, and tests passed.
- Local native validation: 108 tests passed.
- `python3 -m compileall -q src test`: passed.
- `git diff --check`: passed.

## Hardware validation

Real STM32F469DISC validation remains a release gate: internal and SD mnemonic save/load/replace/delete, single-file deletion, full SD erase/format, normal and emergency device wipe, and reboot behavior. If practical, test power loss during save, deletion, and formatting; data loss is allowed by design, but the device must not become stuck or falsely report success.

This PR does not add recovery logic and must not be merged without the hardware validation above.


schnuartz-ai:
Pushed `8585e60` addressing the two remaining review topics (power-loss
limitation wording + scratch-reconciliation hardening). This does **not**
touch the separate SD/VFS I/O error-propagation work.

### 1. Power-loss limitation is now documented, not overclaimed

The `.tmp`/`.old` swap is resilient against interrupted writes but is
**not** fully atomic: FatFs `f_rename()` has a critical section where a
power loss / hard reset can leave a cross-linked directory state, and in
the worst case *both* the previous and the newly written local mnemonic
copy can become unusable. This concerns replacement of an already stored
key only, not normal reading/use.

- `docs/data-storage.md` — new "Replacing a saved mnemonic
  (interrupted-write recovery)" section with the accepted-limitation note
  and "keep an independent backup".
- `docs/security-model.md` — bullet under "Known limitations".
- `src/keystore/flash.py` — corrected `_save_key_file()` /
  `reconcile_scratch_dir()` docstrings; dropped "without ever leaving
  zero valid copies" / "recovers the one state".
- test docstring in `test_save_mnemonic_replace.py` softened; no test
  claims a mid-rename power loss is recoverable.

**Suggested one-line edit to this PR description** (bullet 3), since it
still overclaims:
> - replace mnemonic files through a verified .tmp -> .old -> target swap
>   that is resilient against interrupted writes (FAT rename is not fully
>   atomic — see docs/data-storage.md)

### 2. Scratch reconciliation hardened

**Directory scan** — `reconcile_scratch_dir()` no longer builds a list of
every filename in the SD root. It streams `os.ilistdir()` and retains
only entries in Specter's own key namespace (prefix-matched key files +
the dot-prefixed `.tmp`/`.old` derived from them). Unrelated files are
dropped as the iterator yields and never renamed/deleted. No global
file-count cap; arbitrary unrelated files/thousands of entries stay
supported. `ilistdir` handle still closed before any mutation.

**Oversized `.old` / `.tmp`** — a scratch-named file larger than
`platform.SCRATCH_RECONCILE_MAX_BYTES` (1 MiB; a real encrypted phrase is
< 1 KiB, so this is a ~1000x margin — a normal user never hits it) is
treated as a faulty/tampered card:
- automatic reconciliation **preserves** it, does not start a multi-minute
  secure overwrite, and still reconciles every other entry;
- it is recorded and surfaced under *Manage keys → Review large leftover
  files*, where the user gets name + size + a rough duration estimate and
  chooses **Delete** (existing `secure_delete_file()` path) or **Skip**;
- no path downgrades to plain `os.remove()`.

New constants live in `platform.py` with rationale. `format_size()` /
`secure_delete_duration_estimate()` give `"< 1 minute"` / `"~2-5 minutes"`
style buckets from a deliberately low assumed throughput.

### Tests / validation

- `test/tests_native/test_scratch_reconcile_hardening.py` (16 cases):
  streaming scan retains only Specter entries over a 5000-entry directory,
  unrelated + foreign dotfiles untouched, oversized `.old`/`.tmp`
  preserved while siblings reconcile, unrelated oversized ignored,
  storage-menu review deletes on confirm / preserves on skip, threshold
  and estimate helpers.
- `python3 test/run_native_tests.py`: **158 passed** (was 142).
- `make test` (micropython unix): **14 passed**.
- `python3 -m compileall -q src test`: passed.
- `git diff --check`: clean.
- Existing replacement/recovery tests (target-only, target+`.old`,
  target+`.tmp`, `.old`/`.tmp` without target, failed-swap rollback,
  cross-key recovery) unchanged and green.

Remaining limitation: the mid-`f_rename()` cross-link window is
consciously **not** fixed here — it is now a documented, accepted
limitation with an independent-backup mitigation.

Schnuartz:
@al-munazzim can you review?

Schnuartz:
Did some testings. Looks like its not ready yet, cause of RAM. But we also have a RAM issue (cf. https://github.com/cryptoadvance/specter-diy/issues/398)...

| # | Hardware Test | Status | Result |
|---:|---|:---:|---|
| 1 | Internal Flash — Save → Power Off → Load | ✅ PASS | A test seed was saved to internal flash, the device was completely powered off and restarted, and the stored seed could be loaded again successfully. |
| 2 | Internal Flash — Replace Existing Seed | ✅ PASS | An existing stored seed was replaced. After reloading, only the new seed was available; no stale previous seed appeared. |
| 3 | Internal Flash — Delete → Reboot | ✅ PASS | Two stored keys were deleted. After rebooting the device, both remained deleted and were no longer available. |
| 4 | PIN + Stored Seed | ✅ PASS | PIN-protected access and loading the stored seed were tested successfully. |
| 5 | SD Card — Save → Load | ✅ PASS | A mnemonic was successfully saved to the SD card and could be loaded again. |
| 6 | SD Card — Replace Existing Seed | ✅ PASS | An existing seed stored on the SD card was replaced with a new seed. After reloading, the expected replacement seed was available. |
| 7 | SD Card — Delete Individual File | 🟡 PARTIAL | An individual JSON file was deleted through the new Delete Files flow. PC-side verification is still pending. |
| 8 | Full SD Card Erase + Format | ❌ FAIL | Repeatedly fails on real hardware with `Not enough memory to start the secure erase. No data has been changed - reboot the device and try again.` |
| 9 | Normal Wipe Device | ✅ PASS | Seed, Wallet Desriptors (including naming) deleted from initial flash |
| 10 | Wipe Persistence after Full Power Cycle | ⏳ NOT TESTED | Pending after Test #9. |
| 11 | Emergency Wipe after Exhausting PIN Attempts | ✅ PASS | Seed, Wallet Desriptors (including naming) deleted from initial flash |
| 12 | Power Loss during Internal Seed Replacement | ⏳ NOT TESTED | Optional stress test, not yet performed. |
| 13 | Power Loss during SD Erase/Format | ⛔ BLOCKED | Cannot be meaningfully tested until the full SD erase operation can start successfully. |

al-munazzim:
Thanks for the hardware test table. I pulled the branch locally and the native suite is green for me after installing the test requirements:

```text
Ran 108 tests in 2.020s
OK
```

The full-card format failure you saw is very likely the upfront erase buffer allocation in `platform.SDCard.erase_and_format()`:

```python
chunk_blocks = max(1, (128 * 1024) // block_size)
buf = bytearray(chunk_blocks * block_size)
```

On the device that means one contiguous ~128 KiB allocation before the first destructive write. That is intentionally fail-safe (the current error says no data was changed), but on a fragmented STM32 heap it is probably too ambitious and explains `Not enough memory to start the secure erase`.

I would not merge the format-card part as “ready” until we either reduce that buffer substantially or make it adaptive. Suggested next patch:

* start with a much smaller fixed target, e.g. 16 KiB or 32 KiB, not 128 KiB; or
* try 128 KiB → 64 KiB → 32 KiB → 16 KiB → one block before giving up, always before the first write.

Then please re-run hardware test #8 on the same card immediately after a fresh start and again after a few normal wallet/signing flows, because fragmentation is what matters here. If it still fails even with a <=16 KiB buffer, please capture the free/allocated heap just before `erase_and_format()` allocates the buffer (`gc.mem_free()` / `gc.mem_alloc()` if available in that firmware build) and the card block size from `ioctl(5, None)`.

Safe user-facing remediation meanwhile: if the full-card format path reports this exact “No data has been changed” error, do **not** retry repeatedly in the same session; start fresh first, and if it still fails, remove the card and format/wipe it outside the device. The single-file delete path is independent and should still be usable for individual files.

Given the table, I’d treat this as: core delete/save/wipe logic looks well covered by native tests, but the full-card SD format UI path remains blocked on real-device RAM until the buffer sizing is adjusted and re-tested.

PR #414, opened 2026-09-03
ci: Add reproducible build workflow

Schnuartz:
Adds a reproducible firmware build workflow for Specter DIY.

- Automatically runs when `boot/main/boot.py` changes on `master` (e.g. firmware version bump)
- Can also be triggered manually for a specific commit SHA or tag
- Builds the firmware using the reproducible Docker environment
- Displays the exact message/hash that needs to be signed
- Uploads the unsigned firmware and build artifacts for verification

This provides an additional independent reference build for releases.

### DIFF
--- a/.github/workflows/reproducible-build.yml
+++ b/.github/workflows/reproducible-build.yml
@@ -24,11 +24,12 @@ jobs:
 
     steps:
       - name: Checkout exact source
-        uses: actions/checkout@v4
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
           ref: ${{ inputs.ref || github.sha }}
           submodules: recursive
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Record exact build inputs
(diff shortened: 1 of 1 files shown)


Schnuartz:
@al-munazzim can you review?

al-munazzim (approved):
Reviewed the workflow after the latest repository_dispatch hardening. Verified locally: YAML parses and actionlint 1.7.7 is clean for .github/workflows/reproducible-build.yml.\n\nThe default-branch-only repository_dispatch path plus the reviewed-master-history check addresses the main concern I had about arbitrary branch refs producing official-looking signing requests. Artifacts are clearly marked pre-sign (sha256.presign.txt), which avoids confusing them with final signed release hashes.\n\nOperational follow-up before relying on this for a real release: after merge, do one dry repository_dispatch run against a known master commit/tag and confirm the summary/artifacts contain the expected source commit, submodules, key-set hash, signing message and presign manifest.

Issue #415, opened 2026-09-04
Automate the release signing pipeline (reproducible build + signature finalization)


schnuartz-ai:
## Problem

The Specter DIY release process today is almost entirely manual and undocumented in an executable form. Producing a signed release means, by hand:

1. Building the firmware reproducibly in the Docker environment.
2. Reading the bech32 signing message out of the build log.
3. Getting the authorized vendor signers to sign that message offline.
4. Importing the base64 signatures back into `specter_upgrade.bin` with `upgrade-generator.py import-sig`.
5. Re-hashing the now-signed binaries into `release/sha256.txt`.
6. GPG-signing that manifest offline into `sha256.signed.txt`.

Nothing enforces the ordering or checks the results. This has a few concrete failure modes:

- **Pre-sign vs. final hashes get confused.** `build_firmware.sh hash` writes `release/sha256.txt` *before* any signatures are imported. If that file is published as the release manifest, the hashes describe an unsigned intermediate, not what users actually flash.
- **No automated threshold / signer check.** The production bootloader key set currently requires `bootloader_sig_threshold = 2` and `main_fw_sig_threshold = 2`. Whether the collected signatures actually satisfy that, are cryptographically valid, map to *authorized production keys*, and contain no duplicate signer, is only ever checked by eye.
- **No check that the signed payload is the reviewed one.** Nothing verifies that the tag being released matches the version in the firmware signing message, or that importing signatures didn't change the payload.
- **Release-window friction.** Signers (e.g. for firmware signatures and for the final GPG manifest) are only intermittently available. Because the steps aren't packaged, each one is a fresh manual effort and the whole thing can't be done in a single sitting when a signer *is* available.

`upgrade-generator.py` (message generation + external signature import) and `introspect-binary.py` (cryptographic signature verification, authorized-key mapping, threshold enforcement, embedded-key inspection) already provide the primitives. They just aren't wired into CI.

## Optimization

Wrap the process in **two separate GitHub Actions workflows**, keeping the three trust layers cleanly separated:

```
reproducibility  ->  firmware vendor signatures  ->  final GPG-signed hash manifest
```

**Important constraint: no private signing keys in GitHub Actions.** Firmware signatures stay external/offline, produced by the authorized signers. CI only receives the finished public base64 signatures and verifies/imports them. The final `sha256.txt` is still GPG-signed offline.

### Workflow 1 - Reproducible Firmware Build

[...]

PR #416, opened 2026-09-04
ci: add release finalization & verification workflow

schnuartz-ai:
Part 2 of #415. Companion to #414 (reproducible build workflow).

Adds a manually-dispatched workflow plus a tested signature-verification module
that turn a reviewed release tag + externally produced vendor signatures into a
verified, finalized firmware release.

### Files

- `.github/workflows/release-finalize.yml` — the finalization workflow
- `tools/verify_release_signatures.py` — authoritative signature gate; reproduces
  the on-device `verify_multisig()` policy (vendor-only + `bootloader_sig_threshold`
  for bootloader upgrades; vendor+maintainer + `main_fw_sig_threshold` for
  main-only). Independent of `introspect-binary.py`.
- `tools/tests/test_verify_release_signatures.py` — 16 offline end-to-end
  scenarios using the committed test key set (`bootloader/keys/test`)
- `.github/workflows/release-tooling-tests.yml` — runs those tests on every PR
  touching the release tooling

### No private keys in CI

Firmware signatures are still created offline by the authorized signers. This
workflow only takes the finished base64 signatures as inputs, imports them, and
verifies them. The final `sha256.txt` still has to be GPG-signed offline
afterwards - the summary says so explicitly.

### What the workflow does

1. Refuses to run unless it is the `master` copy of the workflow file
   (`github.workflow_ref`), and `release_tag` matches `^v\d+\.\d+\.\d+(-rc\d+)?$`.
2. Checks out the exact release tag + recursive submodules; fails if the tag
   doesn't exist, the checked-out commit doesn't match it, or the tag commit is
   not an ancestor of `origin/master`.
3. Reads `bootloader_sig_threshold` / `main_fw_sig_threshold` from
   `bootloader/keys/production/pubkeys.c`.
4. Rebuilds the unsigned firmware from that tag in the reproducible Docker image
   (no uploaded binaries).
5. Verifies the `main` payload section version matches the tag.
6. Imports the supplied vendor signatures via `upgrade-generator.py import-sig`.
7. Verifies the signing message is byte-identical before and after import.
8. Runs `tools/verify_release_signatures.py` as the authoritative gate, then
   `introspect-binary.py` as an independent cross-check.
9. Structurally validates the `initial_firmware.bin` ICR (magic, revision,
   self-CRC, payload version) and checks expected production keys are present.
10. Only then writes the final `sha256.txt`, with versioned filenames, and
    uploads the release package + verification artifacts.

### Notes for review

- Four signature inputs are exposed; the workflow hard-fails if the production
  threshold ever exceeds 4.
- The `github.workflow_ref` check is defence in depth, not a hard boundary. A
  default-branch-only trigger (`repository_dispatch`) would make it one - happy
  to switch if preferred.
- The `introspect-binary.py` vendor/maintainer policy gap should be fixed
  upstream in `specter-bootloader` and pulled in via a submodule bump;
  `tools/verify_release_signatures.py` is the in-repo mitigation until then.
- Split from #414 deliberately.


schnuartz-ai:
Pushed hardening changes (89e24d4) after review of the finalization logic against the actual bootloader tooling:

**Fixes**
- **Tag/version check was broken.** It prefix-matched `inputs.release_tag` against the bech32 signing message, but that message starts with the bootloader payload-section name (`b…`) and its first version component is the *bootloader* version, so the check never matched a real release. Now it reads the `main` payload section version from `upgrade-generator.py dump` and compares that to the tag.
- **Tag was not bound to reviewed code.** A tag can point at any commit. Added a `git merge-base --is-ancestor "$TAG_COMMIT" origin/master` gate.
- **Signature policy did not match the device.** `introspect-binary.py` checks vendor+maintainer keys in one combined set and does not restrict a bootloader payload to vendor keys (works today only because the production vendor and maintainer lists are identical). Added a self-contained Python gate that reproduces `verify_multisig()`: `>= bootloader_sig_threshold` cryptographically valid **vendor** signatures for a bootloader upgrade, plus duplicate-signer detection. `introspect-binary.py` is kept as a secondary cross-check only.
- **"valid ICR" was `grep "Found ICR"`**, which only means the INTG magic is present. Replaced with structural validation: magic, supported struct revision, and the ICR's own CRC32, plus payload-version-vs-tag. Full payload-CRC validation needs the memory map and is left to dedicated bootloader tooling.
- Softened the embedded-keys step wording to "present in image" (presence ≠ effective active key set / thresholds).
- Actions pinned to commit SHAs, `persist-credentials: false`, `if-no-files-found: error`.

`actionlint` (with shellcheck) is clean apart from one SC2129 style hint.

**Still open / not solvable in this repo**
- The real fix for the signature-policy gap is in `specter-bootloader`'s `introspect-binary.py` (make it use the same split vendor/maintainer policy as the C bootloader), then bump the submodule here. The in-workflow gate above is a mitigation.
- A proper "the compiled bootloader's effective key set == pubkeys.c" inspector would also belong in `specter-bootloader`.
- These workflows still only run on `workflow_dispatch` / a `boot/main/boot.py` push, so PR CI does not exercise them end to end.

schnuartz-ai:
Good catch — the vendor-policy gate from 89e24d4 was effectively a no-op. Fixed in 01360e6:

- `docker run` was missing `-i`, so `python3 - <<'PY'` got EOF and exited 0 without running the verifier.
- `load_sections` / `parse_sections` are defined in `upgrade-generator.py`, not `core.blsection`, so the import would have failed even with `-i`.

Now the verifier is written to a file and executed in the container (no stdin dependency), and `load_sections` / `parse_sections` are loaded from `upgrade-generator.py` via `importlib` the same way `introspect-binary.py` does it.

**This still needs a real `workflow_dispatch` run** with (a) valid vendor signatures and (b) deliberately wrong / duplicate / non-vendor signatures before merge — CI does not exercise this path. Agreed that the right long-term shape is `tools/verify_release_signatures.py` as a committed, unit-tested module rather than a YAML heredoc; happy to do that as the follow-up if you'd prefer it in this PR.

The `#414` side (bind a manual `ref` dispatch to reviewed master history, so an arbitrary branch can't generate an official signing request) is addressed in that PR's branch.

schnuartz-ai:
Addressed the "which workflow version runs" point (4510cb4):

- New first step refuses to run unless `github.workflow_ref` is exactly `<repo>/.github/workflows/release-finalize.yml@refs/heads/master`. `workflow_dispatch` otherwise executes the workflow file from whatever ref it is dispatched on, so a write-access user could run a tampered `release-finalize.yml` from a branch against a legitimate tag and bypass every in-file check.
- `release_tag` now must match `^v[0-9]+\.[0-9]+\.[0-9]+(-rc[0-9]+)?$`.

Honest about scope: the `workflow_ref` check is defence in depth, not a hard boundary — a branch copy of the file could delete that step too. Making it a real boundary needs one of:
- required review on `.github/workflows/**` (branch protection / ruleset), plus this check, or
- a **default-branch-only trigger** — `repository_dispatch` always runs the default-branch workflow with `GITHUB_REF` = default branch. That's the robust fix but it changes how a release is kicked off (API `POST /repos/{owner}/{repo}/dispatches` with a `client_payload` instead of the Actions UI button). Happy to convert both workflows to `repository_dispatch` if that's the direction you want.

Still open before this is a dependable gate (unchanged from the previous comment):
- the vendor-policy verifier should move to a committed, unit-tested `tools/verify_release_signatures.py` rather than a YAML-generated file
- a real `workflow_dispatch` run with valid + deliberately bad/duplicate/non-vendor signatures
- `v*` tag immutability (protected/immutable tags, or verify signed release tags, or bind an expected commit SHA) — a tag name alone isn't a strong release identity if tags can be moved

schnuartz-ai:
Extracted the signature gate into a tested module (ac41aa6):

- **`tools/verify_release_signatures.py`** — standalone module + CLI. Reproduces the on-device `verify_multisig()` policy: bootloader upgrades need `>= bootloader_sig_threshold` valid **vendor** signatures; main-firmware-only upgrades need `>= main_fw_sig_threshold` from vendor **or** maintainer. Independent of `introspect-binary.py`.
- **`tools/tests/test_verify_release_signatures.py`** — 12 offline end-to-end scenarios using the committed test key set (`bootloader/keys/test/*.pem` + `pubkeys.c`) and in-memory upgrade files. No merge, no real signers, no firmware build needed. `core.recovery.sign_recoverable` produces the exact base64 signatures `import-sig` consumes.
- **`.github/workflows/release-tooling-tests.yml`** — runs those tests on every PR touching the release tooling. **Now green on this PR: 12 passed.**
- `release-finalize.yml` calls the module instead of generating a heredoc.

Scenarios covered (all asserted PASS/FAIL against expectation):

| scenario | expected |
|---|---|
| 2 / 3 distinct vendor sigs (bootloader) | accept |
| 1 vendor sig | reject (below threshold) |
| same vendor key twice | reject (at `import-sig`) |
| vendor + maintainer-only, below vendor threshold | reject |
| 2 vendor + maintainer-only | accept |
| tampered signature byte | reject (invalid signature) |
| signatures over a different message / wrong version | reject |
| valid signature from a key in neither list | reject |
| provided-count ≠ distinct signers (duplicate backstop) | reject |
| main-only: maintainers count toward `main_fw_sig_threshold` | accept |
| main-only below threshold | reject |

Still recommended before merge: the `introspect-binary.py` policy fix belongs upstream in `specter-bootloader` (then a submodule bump), and the `workflow_dispatch` → `repository_dispatch` decision is yours. The signature verification itself is now covered.

schnuartz-ai:
Addressed the key-set invariant point (ed3643f):

- `verify_release_signatures.py` now mirrors the bootloader's `validate_pubkey_set()`: `bootloader_sig_threshold` must be in `[1, #vendor]` and `main_fw_sig_threshold` in `[1, #vendor + #maintainer]`. A threshold of 0 (or one larger than the eligible key count) is an invalid key set the device would refuse to boot with, so reporting "policy satisfied" would be false. Rejected as `invalid key set: …`.
- 4 new tests (16 total, green): zero boot threshold, boot threshold above vendor count, zero main threshold, and a check that the real production `pubkeys.c` (2-of-4 / 2-of-8) is *not* flagged.

Also (the non-blocking packaging point): `specter_upgrade_unsigned_*.bin` moved into `verification/`, so `final-release/sha256.txt` only lists the files that actually carry the vendor signatures (signed upgrade + initial firmware). Summary wording updated to match.

Not changed (agreed non-blocking): `introspect-binary.py` upstream fix + submodule bump, `v*` tag immutability, running the tests inside the pinned Docker image, and the post-merge end-to-end `repository_dispatch` dry run.

schnuartz-ai:
Addressed the trusted-verifier separation (0527059):

The job now does **two checkouts**:
- `trusted/` = master (default branch; `repository_dispatch` always loads the workflow *and* now the verifier from here)
- `source/` = the exact release tag (`defaults.run.working-directory: source`, so every other step is unchanged)

The authoritative gate runs `python3 trusted/tools/verify_release_signatures.py --upgrade source/release/specter_upgrade.bin …` — the verifier code *and* the `bootloader/tools` crypto/parsing it imports come from master, never from the tag. Before running it, the step asserts `sha256(source/…/production/pubkeys.c) == sha256(trusted/…/production/pubkeys.c)` — a key or threshold change is explicitly out of scope for this workflow (also closes the earlier key-rotation follow-up).

Build, `import-sig`, the `introspect-binary.py` cross-check and the ICR/key-embed checks still run from `source/` (they are non-authoritative or are what's actually being released).

Verified the cross-directory invocation in a clean container: `cd $GITHUB_WORKSPACE && python3 trusted/tools/verify_release_signatures.py --upgrade source/release/specter_upgrade.bin …` → ACCEPT on 2 valid vendor sigs, REJECT on 1. `actionlint` clean.

Note: the verifier runs natively via `setup-python` + `pip install cryptography intelhex bitstring click pytest` (pytest only because `core/blsection.py` imports it at module scope), not in the tag's Docker image — so it has zero dependency on tag-controlled code.

schnuartz-ai:
Pinned + hashed the verifier's Python dependencies (176e679):

- `tools/release-requirements.in` (source) + `tools/release-requirements.txt` (`pip-compile --generate-hashes --allow-unsafe --strip-extras`, 12 packages incl. all transitive deps, every one hashed).
- `release-finalize.yml`: the trusted-verifier install step is now `pip install --require-hashes -r trusted/tools/release-requirements.txt` — pinned set, taken from `trusted/` (master).
- `release-tooling-tests.yml`: same lockfile, `pip install --require-hashes -r tools/release-requirements.txt`.

All resolved to wheels, so no compiler/sdist execution on the runner. 16 tests green against the pinned set (verified locally in a clean `python:3.11-slim`, and CI job re-runs on this PR).

`pytest` is in the lockfile only because `bootloader/tools/core/blsection.py` imports it at module scope; noted in the `.in` header. Regen command is in the file header.

Also fixed the PR description (was "12 tests", now 16).

Remaining (both operational, not code — agreed):
- a real end-to-end `repository_dispatch` dry run (positive + negative signature cases) before the workflow is trusted for a production release — best done in a fork/test repo with this branch as the default branch, since `repository_dispatch` intentionally only runs the default-branch version
- confirm `master` branch protection + `v*` tag immutability at the repo-settings level (out of scope for this PR)

Schnuartz (approved):

Schnuartz:
@al-munazzim can you review?

al-munazzim (approved):
Reviewed the release-finalization changes. Verified locally: actionlint 1.7.7 is clean for release-finalize.yml and release-tooling-tests.yml; after initializing recursive submodules and installing the hashed release requirements, tools/tests passed: 16 passed.\n\nThe important security boundaries look right to me now: repository_dispatch/default-branch workflow, trusted verifier from master, source checkout pinned to the release tag, pubkeys.c equality check, vendor-only bootloader signature gate, duplicate/threshold invariants, and hashed Python dependencies.\n\nRemaining items are operational rather than blockers for this PR: run a post-merge repository_dispatch dry run with both valid and deliberately bad signatures in a test/fork setup, and confirm repo settings for master protection and v* tag immutability before production use.

PR #417, opened 2026-09-04
ci: Add PR security & change guard workflow

schnuartz-ai:
## What

Adds one informational GitHub Actions workflow,
`.github/workflows/pr-security-change-guard.yml`, that helps reviewers by
posting **one sticky comment per PR** (also mirrored to
`GITHUB_STEP_SUMMARY`) answering four questions that are easy to miss:

1. **Security-sensitive files** – were files in an explicitly curated
   catalogue of sensitive areas touched (bootloader / f469-disco / boot
   sequence / build & release tooling / CI / keystore / mnemonic & seed
   entry / RNG / wallet-PSBT-descriptor / **xpub export & authorization**
   / message signing / blinding keys / host transports / core
   coordinator / reproducible-build metadata)? It deliberately does
   **not** flag every `src/` change.
2. **Production signing configuration** – if the `bootloader` submodule
   pointer moved, it loads `keys/production/pubkeys.c` from the old and
   the new bootloader commit and compares the vendor + maintainer public
   keys and both signature thresholds (`bootloader_sig_threshold`,
   `main_fw_sig_threshold`). Reports `✅ unchanged`, a prominent
   `⚠️ changed` block (old/new thresholds, old/new key-set SHA256,
   added/removed key owners + fingerprints), or `❓ unable to verify`.
3. **Submodule pointer changes** – top level (`bootloader`, `f469-disco`)
   **and one nested level**, discovered by reading each submodule's
   `.gitmodules` (so new nested submodules – e.g. `libs/common/embit` –
   are picked up automatically). Detects moved **and added/removed**
   pointers, with `old...new` compare links.
4. **Changes since last approval** – finds the most recent `APPROVED`
   review (recovering an approval that was later **dismissed as stale**
   via the PR timeline), compares its `commit_id` with the current head,
   and lists commits / files / `+/-` lines added since, re-highlighting
   security-sensitive and submodule changes. `ℹ️ No previous approval
   found` is treated as normal, not an error.

## Not blocking – but fail-closed

Every finding is informational. The workflow never fails the check,
never blocks merge, never dismisses reviews and never touches
branch-protection rules.

For its **own** reporting it fails closed: when it cannot determine
something (API error, unparseable `pubkeys.c`, unresolvable pointer) it
prints a distinct `❓ … manual review required` instead of a green `✅`.

## Fork-PR safety

Uses `pull_request_target` **only** to obtain a token that can update the
sticky comment. It:

* has no `actions/checkout` step at all,
* never runs any script, action or tool originating from the PR,
* reads all PR data exclusively through authenticated GitHub REST API
  calls,
* runs only `actions/github-script`, **pinned to a full commit SHA**,
  with code from the base branch (this file),
* has no `run:` steps, so no PR-controlled string is ever interpolated
  into a shell command; externally sourced strings (file names, key
  owner comments) are Markdown-escaped before display.

Because no PR-controlled code is executed, the writable token is never
exposed to attacker-controlled code.

Permissions: `contents: read`, `pull-requests: write`.
Trigger: `pull_request_target` [opened, reopened, synchronize,
ready_for_review] only. `pull_request_review` is deliberately not used —
on a fork PR GitHub runs it with a read-only token, so the comment
update would fail; `synchronize` covers the case that matters (new
commits after an approval). The sticky comment is an issue comment
(matched by marker **and** bot author), which does not re-trigger the
event, so there is no loop.

## Known limitations

* `pull_request_target` always runs the workflow version on the base
  branch, so this cannot fully self-test on its own PR. It has been
  YAML- and JS-syntax-checked and its parsers / branches exercised
  offline against the real `pubkeys.c` and `.gitmodules` and against
  mocked API responses (clean PR, threshold change, unparseable keys,
  nested submodule added, listFiles failure, dismissed approval).
* The bootloader `pubkeys.c` parser is regex-based and intentionally
  strict; a future refactor of that file will trip the `❓ unable to
  verify` path rather than silently misreport.
* Commit-range comparisons for "changes since approval" rely on the
  approved commit still being reachable; a force-push that removed it
  yields `❓ manual review`.

## Review round 3 (commit `1a8bc4b`)

* Nested-submodule scan is now genuinely fail-closed: reading a parent
  `.gitmodules` treats only a 404 as "no submodules"; a 403/429/5xx marks
  the scan incomplete (`❓`) instead of being swallowed and reported as
  "unchanged".
* Removed the `pull_request_review` trigger (see above).
* Timeline API failure while checking for a dismissed approval no longer
  degrades to "No previous approval found" — it becomes a distinct `❓`
  when there is also no standing approval.
* `pathState` now tries every candidate head ref before concluding
  "absent".
* Pinned `actions/github-script` SHA corrected in the comment to `v7.1.0`
  (the SHA was right, the version label was wrong).

Offline scenario coverage extended to: nested `.gitmodules` API error,
and dismissed-approval + timeline-down.

## Review round 4 (commit `b6cba92`)

* **Blocker fixed:** "changes since last approval" now checks the compare
  status. `compareCommitsWithBasehead` returns `200 diverged` / `behind`
  when the branch was rebased / force-pushed after the approval (the
  approved commit is no longer an ancestor of HEAD). Only status `ahead`
  is shown as a normal post-approval file list; `diverged` / `behind` /
  API error all render `❓ … rebased / force-pushed / rewritten … manual
  review required`.
* `.gitmodules` 404 handling: a 404 is treated as "no submodules" only
  after `getCommit` confirms the parent commit resolves; an unreachable
  ref/repo now marks the nested scan incomplete.

Offline scenarios added: rebased-after-approval (`diverged`), compare API
down, `.gitmodules` 404 with commit present vs. commit also unreachable.

### DIFF
--- a/.github/workflows/pr-security-change-guard.yml
+++ b/.github/workflows/pr-security-change-guard.yml
@@ -74,9 +74,11 @@ jobs:
     steps:
       - name: Build & publish report
         # Pinned to a full commit SHA. This SHA is the actions/github-script
-        # v7.1.0 release (also where the moving `v7` tag points). Update the
-        # SHA and the trailing comment together when bumping.
-        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        # v8.0.0 release (also where the moving `v8` tag points). v8 runs on
+        # the Node 24 runtime; GitHub is removing Node 20 from hosted runners
+        # in 2026, so v7 must not be used for a newly introduced workflow.
+        # Update the SHA and the trailing comment together when bumping.
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
         with:
           # Default GITHUB_TOKEN. Scope is limited by `permissions:` above.
           script: |
@@ -187,6 +189,29 @@ jobs:
             const short = (sha) => (sha ? sha.substring(0, 12) : '(none)');
 
             // ---------------------------------------------------------------
+            // Baseline for "what did THIS PR change?"
+            // ---------------------------------------------------------------
+            // `pr.base.sha` is the *current* tip of the base branch, which may
+            // have moved on with unrelated commits since the PR was opened.
+            // Comparing submodule / signing config against it would attribute
+            // those unrelated base-branch changes to this PR. The correct
+            // baseline is the merge-base of base...head. Fall back to
+            // `pr.base.sha` only if the compare API is unavailable.
+            let mergeBaseSha = baseSha;
+            try {
+              const mb = await github.rest.repos.compareCommitsWithBasehead({
+                owner, repo, basehead: `${baseSha}...${headSha}`,
+              });
+              if (mb.data.merge_base_commit && mb.data.merge_base_commit.sha) {
+                mergeBaseSha = mb.data.merge_base_commit.sha;
+              }
+            } catch (e) {
+              note(`Could not determine the base/head merge-base (${e.message}); ` +
+                   `comparing against the base branch tip \`${short(baseSha)}\` instead — ` +
+                   `unrelated base-branch changes may be misattributed to this PR.`);
+            }
+
+            // ---------------------------------------------------------------
             // Security-sensitive path catalogue
             // (derived from an inspection of the current repository layout)
             // ---------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


Schnuartz (approved):
Think this just makes reviewing of PRs easier.

Schnuartz:
@al-munazzim can you review?

al-munazzim (approved):
Reviewed the PR security/change guard workflow after the latest fail-closed changes. Verified locally with actionlint 1.7.7: .github/workflows/pr-security-change-guard.yml is clean.\n\nI like that this stays informational-only, uses pull_request_target without checkout/run of PR code, escapes PR-controlled strings, compares sensitive/submodule state against the merge-base, handles renamed files, treats API caps/merge-base failures as manual review, and flags rewritten-history-after-approval instead of presenting a misleading clean diff. No blocker from me.

